PR #5, opened 2018-03-06
Fixed missing dependency

tboerger:
Currently the Gopkg.lock file is missing the dependency
github.com/alecthomas/template used by some drivers. With this commit
the dependencies should be clear.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -32,6 +32,15 @@
   version = "v0.4.7"
 
 [[projects]]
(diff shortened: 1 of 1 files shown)


bradrydzewski:
I pushed 62d78b7a6e991d5acf7bf6c9ad98133f23d325c0 and replaced `github.com/alecthomas/template` with `text/template`. Looks like goimports watch automatically importing the wrong package. Good catch.

PR #7, opened 2018-03-06
Make cloud-init/userdata configurable

tboerger:
Currently the userdata is staticaly compiled in, with this change the
admin is able to customize the userdata or cloud-init content, the
content can be provided via path, base64 encoded string or a plain
string.

### DIFF
--- a/drivers/amazon/create.go
+++ b/drivers/amazon/create.go
@@ -8,12 +8,12 @@ import (
 	"bytes"
 	"context"
 	"encoding/base64"
+	"text/template"
 	"time"
 
 	"github.com/drone/autoscaler"
 	"github.com/drone/autoscaler/drivers/internal/scripts"
 
-	"github.com/alecthomas/template"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/rs/zerolog/log"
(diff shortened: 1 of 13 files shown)


drivers/amazon/create.go
```diff
@@ -10,9 +10,10 @@ import (
 	"encoding/base64"
 	"time"
 
-	"github.com/alecthomas/template"
 	"github.com/drone/autoscaler"
+	"github.com/drone/autoscaler/drivers/internal/scripts"
 
+	"github.com/alecthomas/template"
```
bradrydzewski: use text/template :)

drivers/amazon/provider.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	"github.com/drone/autoscaler"
 
+	"github.com/alecthomas/template"
```
bradrydzewski: text/template

bradrydzewski:
looks good, can you replace `alecthomas/template` with `text/template`?  I had intended to use `text/template`, but goimports chose the wrong package.

drivers/amazon/create.go
tboerger: Just followed your example :D

drivers/amazon/provider.go
tboerger: Just followed your example :D

tboerger:
Replaced with text/template now.

PR #8, opened 2018-03-08
add cloud-init path config, allow base64 encoded user-data

tboerger:


config/config.go
```diff
@@ -99,12 +102,13 @@ type (
...
-			Datacenter string
-			Image      string
-			UserData   string `split_words:"true"`
-			SSHKey     int
-			Token      string
-			Type       string
+			Datacenter   string
+			Image        string
+			UserData     string `split_words:"true"`
+			UserDataPath string `env:"DRONE_HETZNERCLOUD_USERDATA_PATH"`
```
bradrydzewski: `DRONE_HETZNERCLOUD_USERDATA_FILE` to match `DRONE_HETZNERCLOUD_USERDATA`

config/config.go
```diff
@@ -73,20 +74,22 @@ type (
...
+			Size         string
+			Tags         []string
 		}
 
 		Google struct {
 			Zone         string `default:"us-central1-a"`
 			MachineType  string `split_words:"true" default:"n1-standard-1"`
 			MachineImage string `split_words:"true" default:"ubuntu-1510-wily-v20151114"`
 			UserData     string `split_words:"true"`
+			UserDataPath string `env:"DRONE_GOOGLE_USERDATA_PATH"`
```
bradrydzewski: `DRONE_GOOGLE_USERDATA_FILE`

config/config.go
```diff
@@ -73,20 +74,22 @@ type (
...
-			UserData string `split_words:"true"`
-			Region   string
-			SSHKey   string
-			Size     string
-			Tags     []string
+			Token        string
+			Image        string
+			UserData     string `split_words:"true"`
+			UserDataPath string `env:"DRONE_DIGITALOCEAN_USERDATA_PATH"`
+			Region       string
```
bradrydzewski: `DRONE_DIGITALOCEAN_USERDATA_FILE`

config/config.go
```diff
@@ -73,20 +74,22 @@ type (
...
-			Image    string
-			UserData string `split_words:"true"`
-			Region   string
-			SSHKey   string
-			Size     string
-			Tags     []string
+			Token        string
+			Image        string
+			UserData     string `split_words:"true"`
+			UserDataPath string `env:"DRONE_DIGITALOCEAN_USERDATA_PATH"`
```
bradrydzewski: I think we might need to remove `split_works` and specify the full environment variable name.

tboerger: I have not defined split words for this variable :(

bradrydzewski:
for loading userdata from file, lets use an `_FILE` variable like what we are doing with amazon (I just added an example). In terms of base64 the existing code is good to go.

tboerger:
I think all your comments should be applied now.

PR #9, opened 2018-03-14
[WIP] packet.net autoscaler implementation

krasi-georgiev:
TODO 
- [x] tests
- [x] ~how to handle when the docker provisioning fails? maybe should reread the env variable and update the device?~ - opened another PR for failed docker installations. https://github.com/drone/autoscaler/pull/10
- [ ] userdata doesn't fully install docker.
- [ ] update the docs.


config/config.go
```diff
@@ -111,5 +111,18 @@ type (
...
+			APIKey       string
+			Facility     string
+			Plan         string
+			OS           string
+			ProjectID    string `split_words:"true"`
+			Tags         []string
+			SSHKey       string
+			UserData     string `envconfig:"DRONE_PACKET_USERDATA"`
+			UserDataFile string `envconfig:"DRONE_PACKET_USERDATA_FILE"`
+			Hostname     string `envconfig:"DRONE_PACKET_HOSTNAME"`
```
tboerger: You can drop this envconfig, this is the same env variable name by default.

drivers/packet/create.go
```diff
@@ -5,15 +5,89 @@
...
 
 import (
+	"bytes"
 	"context"
+	"fmt"
+	"strings"
+	"time"
 
 	"github.com/drone/autoscaler"
+	"github.com/packethost/packngo"
```
tboerger: This package have to be added to Gopkg.toml

drivers/packet/create.go
```diff
@@ -5,15 +5,89 @@
...
+	}
+
+	d, _, err := p.client.Devices.Create(&cr)
+	if err != nil {
+		return nil, err
+	}
+	id := d.ID
+
+	var IPAddress string
+	for i := 0; i < 100; i++ {
```
tboerger: Maybe use the same block as DigitalOcean, e.g. https://github.com/drone/autoscaler/blob/master/drivers/digitalocean/create.go#L79-L114?

drivers/packet/option.go
```diff
@@ -56,7 +56,7 @@ func WithSSHKey(sshkey string) Option {
 }
 
 // WithTags returns an option to set the image.
-func WithTags(tags ...string) Option {
```
tboerger: The tags have been intentionally done like that :)

krasi-georgiev: thanks for the pointer , updated the PR 

`tags ...string` on this one what is the reason to format it like this?

drivers/packet/create.go
krasi-georgiev: it was already there

krasi-georgiev:
@zsmith928  the device creation works . do you provide an easy way to provision docker on a new device?

krasi-georgiev:
for the docker installation I am thinking to run the official installation depending on the OS image selected for the device provisioning.
https://docs.docker.com/install/


btw @bradrydzewski , @tboerger  have you considered using docker-machine ? 
this way you just need to consume docker-machine as a library with the providers plugins. 
this will reduce the maintenance on your side as many cloud providers support and maintain their own docker-machine plugin.
http://github.com/packethost/docker-machine-driver-packet for example.

drivers/internal/userdata/userdata.go
```diff
@@ -28,8 +30,19 @@ func Parse(text string) *template.Template {
 	)
 }
 
-// T is the default userdata template.
-var T = Parse(`#cloud-config
+// DetectUserdata returns the docker provisioning userdata script based on the OS string
+func DetectUserdata(os string) (*template.Template, error) {
+	switch {
```
bradrydzewski: lets move this to a separate pull request and keep this PR focused just on the packet implementation. This will likely always fail for amazon, since the image is formatted ami-80ca47e6 and does not contain amazon.

krasi-georgiev:
thanks for the pointers.  any comments about using docker machine as a library?

bradrydzewski:
> any comments about using docker machine as a library?

Docker machine lacks some abstractions that would have made embedding a more viable option at the start. I now prefer our implementation, which uses cloud-init, over the docker-machine implementation which uses ssh and is tightly coupled to docker at the driver-level. This gives us more flexibility to support other container runtimes and operating systems (windows). So no immediate plans to adopt machine.

drivers/packet/option.go
tboerger: to call it like `packet.WithTags("a", "b", "c")` instead of `packet.WithTags([]string{"a", "b", "c"})`.

bradrydzewski:
Regarding this todo: 

> how to handle when the docker provisioning fails? maybe should reread the env variable and update the device?

Checking the docker installation is handled separately, outside of the driver code. So this is not something that needs to be considered in the packet.net driver implementation.

Pending unit tests and my feedback regarding user-data, I think the implementation is looking great and can be merged.

drivers/packet/option.go
bradrydzewski: For consistency with other drivers lets use `tags ...string` in this initial packet implementation. I am open to changing the function signature to `tags []string` but I would recommend handling this change in a separate pull request, and updating all drivers :)

drivers/packet/provider.go
```diff
@@ -43,20 +45,23 @@ func New(opts ...Option) autoscaler.Provider {
 		p.facility = "ewr1"
 	}
 	if p.os == "" {
-		p.os = "ubuntu_16_04"
+		p.os = "ubuntu_17_10"
```
bradrydzewski: my preference is that we default to `ubuntu_16_04` since it is designated a long term support release, vs 17.10 which I believe only has a 9 month support period.

krasi-georgiev: the cloud-config script is not deploying docker properly  so I tried few different OS versions.
I am on the packet slack so once this is resolved will update the PR one last time.

bradrydzewski: got it ... also I just checked and 18.04 is released in 5 weeks. So we can just pick whatever works in the short term, and then update to 18.0.4 when available

bradrydzewski: the cloud-init script is working on gcp/hetzner/amazon on ubuntu 16.04, so it is interesting it does not work on the Packet image. Anything in the cloud-init logs or `journalctl -fu docker.service` that might provide insight?

krasi-georgiev:
@bradrydzewski it would be super helpful if you could join the packet slack channel to give some input for the `cloud-config` as we can't make the default one work there.

When was the last time and which provider was this confirmed to work ok?

tboerger:
Let's see if packet got some free credits, than I can do some debugging of the cloud init.

krasi-georgiev:
@tboerger ping me privately and I will give you access.
#prometheus-dev , krasi

bradrydzewski:
Note that Digital Ocean uses a custom cloud-init file, since docker is already installed on the default image. So if necessary, we can have a separate userdata script just for Packet. It could be cloud-init or just a simple script; doesn't really matter as long as docker is installed, certs installed and tls configured.

bradrydzewski:
any word from the packet folks?

krasi-georgiev:
the apt module is not added to their images. still waiting.

tboerger:
The initial addition had been done, but it was buggy and it was missing the https transport for apt. Since than no further response

bradrydzewski:
@tboerger any updates from packet?

krasi-georgiev:
just pinged manny form packet and he thinks it should be ok now but someone needs to test it and confirm.

tboerger:
Just gave it another try, the Ubuntu 16.04 image lacks the https apt transport, so the cloud-init install still fails.

krasi-georgiev:
I just talked to @mmlb again and he won't be able to look into this anytime soon.
as a workaround he suggest that we can provide our own image as a workaround to unblock the PR.
Another PR would be using the official one provided by packet when it becomes available. 


@tboerger  here is a short info how to provide a custom image(copy and paste from the chat with @mmlb), but probably need to ping @mmlb for more info

```
image_repo=some-publically-accessible-git-lfs-repo
image_tag=any-git-commitish
```
so we will clone that repo
checkout the $image_tag
and git checkout the tag, smudge/clean filters are setup so that git-lfs will fetch the files if necessary
then we expect the following files in the repo root:
```
kernel.tar.gz
initrd.tar.gz
modules.tar.gz
image.tar.gz
```

mmlb:
@krasi-georgiev plop this into the user-data and you can use this image until we get the image deployed to prod.

```
#image_repo=https://github.com/packethost/packet-images.git
#image_tag=de2157faee53a67f76591dd2e99318a0ea110a5d
```

bradrydzewski:
@mmlb thanks for the update. When @tboerger is back from vacation we'll see if he can run a quick test and if successful, we can get this merged.

krasi-georgiev:
seems this is not going anywhere and we have moved into a different direction so won't have time to finish this so closing.

Feel free to clone the repo and take over.
Sorry.

krasi-georgiev:
yep seems that way :) sorry

krasi-georgiev:
please clone the branch as I will probably delete it at some point.

tboerger:
Looks like the cloud-init works now... i will give it a try

krasi-georgiev:
Nice, good luck! 
I am unsubscribing now, so ball is in your court :+1:

PR #10, opened 2018-03-15
run the installer for staging servers at bootup

krasi-georgiev:


bradrydzewski:
Yes, good idea. I'd like to propose we place this in a function called `Restart` in the `engine.installer`. This can then be invoked in `engine.Start`:

```diff
func (e *engine) Start(ctx context.Context) {
+	e.installer.Restart()
+
	var wg sync.WaitGroup
	wg.Add(5)
```

Thanks!

krasi-georgiev:
Having this as a function is logical.
I wanted to keep the Start function clean and thought that since it is related to the installation should sit in the install function logic. If more of these workarounds start poping up the Start function might become a bit overcrowded. Don't mind moving it if you prefer to have it in Start.

bradrydzewski:
looks good, thanks!

PR #11, opened 2018-03-27
Start a drone agent for each parallel process in the build

fhp:
Use the number of processes for a build to calculate the required number of Drone agents.


fhp:
Closed this branch as Drone kept building the same old git-version over and over.

PR #13, opened 2018-03-27
 Start a drone agent for each parallel process in the build

fhp:
Use the number of processes for a build to calculate the required number of Drone agents.

### DIFF
--- a/engine/planner.go
+++ b/engine/planner.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"sort"
 	"time"
+	"os"
 
 	"github.com/drone/autoscaler"
 	"github.com/drone/autoscaler/limiter"
@@ -207,6 +208,9 @@ func (p *planner) mark(ctx context.Context, n int) error {
 // helper function returns the number of pending and
 // running builds in the remote Drone installation.
 func (p *planner) count(ctx context.Context) (pending, running int, err error) {
+	if os.Getenv("ENABLE_MATRIX_CALC") == "true" {
+		return p.countMatrix(ctx)
+	}
 	activity, err := p.client.BuildQueue()
 	if err != nil {
 		return pending, running, err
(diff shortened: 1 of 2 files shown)


bradrydzewski:
Thanks, I discussed something similar with @tboerger. In the 1.0 release we will be replacing the matrix capability with something called triggers [1]. For this reason, I would rather put matrix-based calculations behind a flag, as shown below, so it can be easily removed from the code once triggers are available:

```
func (p *planner) capacity(ctx context.Context) (capacity, count int, err error) {
+	if os.Getenv("ENABLE_MATRIX_CALC") == "true" {
+		return p.countMatrix(ctx)
+	}
	servers, err := p.servers.List(ctx)
	if err != nil {
		return capacity, count, err
	}
	for _, server := range servers {
		switch server.State {
		case autoscaler.StateStopped:
			// ignore state
		default:
			count++
			capacity += server.Capacity
		}
	}
	return
}

+func (p *planner) countMatrix(ctx context.Context) (pending, running int, err error) {
+	activity, err := p.client.BuildQueue()
+	if err != nil {
+		return pending, running, err
+	}
+	for _, activity := range activity {
+		build, err := p.client.Build(activity.Owner, activity.Name, activity.Number);
+		if err != nil {
+			return pending, running, err
+		}
+
+		for _, process := range build.Procs {
+			if process.State == drone.StatusPending {
+				pending++
+			} else {
+				running++
+			}
+		}
+	}
+	return
+}
```

[1] https://discourse.drone.io/t/roadmap-for-the-1-0-release/1723

engine/planner.go
```diff
@@ -212,10 +212,17 @@ func (p *planner) count(ctx context.Context) (pending, running int, err error) {
...
+		if err != nil {
+			return pending, running, err
+		}
+
+		for _, process := range build.Procs {
+			if process.State == drone.StatusPending {
+				pending++
+			} else {
+				running++
+			}
```
bradrydzewski: I think this could end up counting matrix steps that are in a finished state. You might want to do something like this instead: 

```
switch process.State {
case drone.StatusPending:
  pending++
case  drone.StatusRunning:
  running++
}
```

tboerger:
I would love to see it implemented like brad said.

engine/planner_test.go
```diff
@@ -337,6 +314,59 @@ func TestPlan_ShutdownIdle(t *testing.T) {
...
 
+// This test verifies that if the server capacity is
+// less than the pending count, including matrix builds,
+// and the server capacity is < the pool maximum,
+// additional servers are provisioned.
+//
+// However, this test should be called with the environment
+// variable "ENABLE_MATRIX_CALC" set to "true". As this is 
+// not easy to mock, it is currently disabled.
+// func TestPlan_MatrixMoreCapacity(t *testing.T) {
```
tboerger: Just set the env variable within this test case? Than you are able to test both scenarios.

bradrydzewski:
awesome, thanks for making the adjustments!

fhp:
Thanks for merging :+1:

PR #15, opened 2018-04-17
add keepalive functionality to booting agents

genisd:
This helps with tcp loadbalancers which would otherwise timeout the
connection. Having the connection timeout is problematic because the
drone-agent does not recognize that it lost the connection.

This keepalive features was added in drone-agent 0.8.4 hence we bump
the default version to 0.8.4

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -35,10 +35,12 @@ type (
 		}
 
 		Agent struct {
-			Host        string
-			Token       string
-			Image       string `default:"drone/agent:0.8"`
-			Concurrency int    `default:"2"`
+			Host             string
+			Token            string
+			Image            string        `default:"drone/agent:0.8"`
+			Concurrency      int           `default:"2"`
+			KeepaliveTime    time.Duration `default:"6m"`
+			KeepaliveTimeout time.Duration `default:"30s"`
 		}
 
 		HTTP struct {
--- a/engine/install.go
+++ b/engine/install.go
@@ -25,6 +25,8 @@ type installer struct {
 	image  string
 	secret string
 	server string
+	keepaliveTime time.Duration
+	keepaliveTimeout time.Duration
 
 	servers autoscaler.ServerStore
 	client  clientFunc
@@ -133,6 +135,8 @@ poller:
 				fmt.Sprintf("DRONE_SERVER=%s", i.server),
 				fmt.Sprintf("DRONE_MAX_PROCS=%v", instance.Capacity),
 				fmt.Sprintf("DRONE_HOSTNAME=%s", instance.Name),
+				fmt.Sprintf("DRONE_KEEPALIVE_TIME=%s", i.keepaliveTime),
+				fmt.Sprintf("DRONE_KEEPALIVE_TIMEOUT=%s", i.keepaliveTimeout),
 			},
 			Volumes: map[string]struct{}{
 				"/var/run/docker.sock": {},
--- a/engine/install.go
+++ b/engine/install.go
@@ -11,6 +11,9 @@ import (
 	"io/ioutil"
 	"sync"
 	"time"
+	"regexp"
+	"strconv"
+	"os"
 
 	"github.com/drone/autoscaler"
 	"github.com/rs/zerolog/log"
@@ -125,6 +128,22 @@ poller:
 		Str("image", i.image).
 		Msg("create agent container")
 
+	keepaliveTimeEnv := os.Getenv("DRONE_AGENT_KEEPALIVE_TIME")
+	if keepaliveTimeEnv != "" {
+		logger.Debug().
+			Str("value", keepaliveTimeEnv).
+			Msg("DRONE_AGENT_KEEPALIVE_TIME override set")
+		i.keepaliveTime = ParseMinuteSecondTimeDuration(keepaliveTimeEnv)
+	}
+
+	keepaliveTimeoutEnv := os.Getenv("DRONE_AGENT_KEEPALIVE_TIMEOUT")
+	if keepaliveTimeEnv != "" {
+		logger.Debug().
+			Str("value", keepaliveTimeoutEnv).
+			Msg("DRONE_AGENT_KEEPALIVE_TIMEOUT override set")
+		i.keepaliveTimeout = ParseMinuteSecondTimeDuration(keepaliveTimeoutEnv)
+	}
+
 	res, err := client.ContainerCreate(ctx,
 		&container.Config{
 			Image:        i.image,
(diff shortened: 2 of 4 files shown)


genisd:
@tboerger ah I didn't know that. Will do thanks!
It's still wip, I'll let you guys know once it's ready

genisd:
Is this the best place to put this? It didn't really `feel` clear to me what the best location was for this.

Also do you know where the docs for are for the autoscaler? If possible I would like to create a PR for this as well

config/config.go
```diff
@@ -35,10 +35,12 @@ type (
...
 		Agent struct {
-			Host        string
-			Token       string
-			Image       string `default:"drone/agent:0.8"`
-			Concurrency int    `default:"2"`
+			Host             string
+			Token            string
+			Image            string `default:"drone/agent:0.8"`
+			Concurrency      int    `default:"2"`
+			KeepaliveTime    string `default:"6m"`
```
tboerger: Both types should be time.Duration

genisd: Done. I was having a bit of trial and error and some missunderstandings.
I'll try this PR (`f252e2a`) on our environment now

genisd: It doesn't build in our environment. I think I had this before, a `git commit --amend` and force push actually yields the previous commit beeing actually built/tested.

I had tried time.Duration before and got the same error infact hence I changed it to strings.
I believe time.Duration if casted to a string get's printed out as an int in nanoseconds (though I surely can fix that in the print statement and print it as say 300s (in seconds basically)).

I'll carry on tomorrow I guess, thanks for the quick feedback through! Appreciated.

The error is as follows: 

```
=== RUN   TestLoad
--- FAIL: TestLoad (0.00s)
	load_test.go:138: json: cannot unmarshal string into Go struct field .KeepaliveTime of type time.Duration
FAIL
```

genisd: I changed it to nanoseconds in the load_test. That should build successfully

genisd:
turns out I was wrong.
```
package main

import (
	"fmt"
	"time"
)

func main() {
	keepaliveTime := time.Duration(360000000000)
	keepaliveTimeout := time.Duration(30000000000)
	fmt.Println(fmt.Sprintf("DRONE_KEEPALIVE_TIME=%s", keepaliveTime))
	fmt.Println(fmt.Sprintf("DRONE_KEEPALIVE_TIMEOUT=%s", keepaliveTimeout))
}
```

prints 
```
DRONE_KEEPALIVE_TIME=6m0s
DRONE_KEEPALIVE_TIMEOUT=30s
```

not sure if `6m0s` will work for the `DRONE_KEEPALIVE_TIME`. I'll let you know :-)

genisd:
This has been running well for us.
Only thing is, is it okay to have this enabled by default?
It's not ENV override-able, should I build that?

genisd:
I've set `DRONE_KEEPALIVE_TIME` to 1 minute to reduce the amount of time for it to detect that the connection is dead.

genisd:
don't merge this. with `1m` this is lower than the serverside enforced default of 5m
default *has to be* >=6m to sync up with serverside default
probably has to be configurable too. it's kinda neccessarry

we're still playing around with this a little ourselves on google kubernetes with it's loadbalancer

genisd:
set the default to 6m / 30s again as suggested in the docs (and enforced by the grpc server side)

yellowmegaman:
This totally fixed our issues (running about 10-11 hours)

https://discourse.drone.io/t/autoscaler-gce-drone-agents-require-restart-after-idling/1964

genisd:
this is good to go as far as i'm concerned. the only "flaw" in the code is that one could pass in stuff like `1h1m1s` and the `1h` would not get parsed and get silently ignored. Imho not a very big flaw :-)

engine/install.go
```diff
@@ -191,3 +214,26 @@ func (i *installer) errorUpdate(ctx context.Context, server *autoscaler.Server,
 	}
 	return err
 }
+
+func ParseMinuteSecondTimeDuration(str string) time.Duration {
```
bradrydzewski: fwiw there is a duration parser in the standard library https://godoc.org/time#ParseDuration

but in this case the durations are already being parsed and set in the Config structure, so you should not need to re-parse this information.

engine/install.go
```diff
@@ -123,6 +128,22 @@ poller:
 		Str("image", i.image).
 		Msg("create agent container")
 
+	keepaliveTimeEnv := os.Getenv("DRONE_AGENT_KEEPALIVE_TIME")
+	if keepaliveTimeEnv != "" {
```
bradrydzewski: this data was already parsed when the `Config` structure was loaded by the system, and should not need to be re-parsed. Also setting `i.keepaliveTime` this way can cause a race condition, and could result in the program crashing.

config/config.go
```diff
@@ -35,10 +35,12 @@ type (
...
-			Host        string
-			Token       string
-			Image       string `default:"drone/agent:0.8"`
-			Concurrency int    `default:"2"`
+			Host             string
+			Token            string
+			Image            string        `default:"drone/agent:0.8"`
+			Concurrency      int           `default:"2"`
+			KeepaliveTime    time.Duration `default:"6m"`
+			KeepaliveTimeout time.Duration `default:"30s"`
```
bradrydzewski: note that when you declare these variables here, they are automatically being parsed from the environment when the system loads. In this case, the system will look for environment variables using `DRONE_AGENT_KEEPALIVETIME` and `DRONE_AGENT_KEEPLIVETIMEOUT` (no space). Note that you can customize the environment variable names. See some of the other attributes in the Config object for examples or refer to https://github.com/kelseyhightower/envconfig

genisd: Thank you for all the feedback.
I was a bit puzzled how certain env vars did get parsed, your comments clear up a lot here.

I hope to have some time Thursday/Friday to apply your feedback.

yellowmegaman:
Still working perfectly BTW 😂

tboerger:
> Still working perfectly BTW 😂

But still missing changes, so it won't get merged like it is.

frebib:
@genisd Any chance on this getting fixed so we can have it merged please?

tboerger:
Could be nice to have that for Drone 0.8, but since Drone 0.9 will drop the GRPC communication I'm pretty sure we won't need that anymore (for Drone >= 0.9).

frebib:
Oh I wasn't aware that 0.9 replaced GRPC (it's kind of hard to follow that without a commit log :roll_eyes:). Currently for us, Autoscaler causes more issues than it solves with the constant disconnecting of agents :/ I hope 0.9 resolves the issues

tboerger:
With 0.9 we will switch back to simple RESTless APIs, so I'm pretty sure the connection will improve because of solid battle-tested infrastructure :)

bradrydzewski:
merged via #27

PR #16, opened 2018-04-19
ignore stopped/shutdown/stopping/error from current capacity

genisd:
the following states shutdown/stoppping/error should not count as available capacity.
so we ignore these as well next to stopped

### DIFF
--- a/engine/planner.go
+++ b/engine/planner.go
@@ -258,6 +258,12 @@ func (p *planner) capacity(ctx context.Context) (capacity, count int, err error)
 		switch server.State {
 		case autoscaler.StateStopped:
 			// ignore state
+		case autoscaler.StateShutdown:
+			// ignore state
+		case autoscaler.StateStopping:
+			// ignore state
+		case autoscaler.StateError:
+			// ignore state
 		default:
 			count++
 			capacity += server.Capacity


bradrydzewski:
This logic is intentional and removing would be dangerous. The autoscaler counts these states toward your capacity to prevent Drone from creating an unbounded number of servers which could result in huge bills. Consider this scenario:

1. Drone creates an AWS server
2. The setup fails (the apt server is unavailable)
3. The server still exists but is placed in an error state
4. Drone repeats this process infinitely, creating hundreds of servers.

So while I appreciate the PR this is not something we would consider merging. You are definitely welcome to remove this logic in your fork, but just be careful, especially when removing the error state.

genisd:
We're running with Google preemtible instances (in AWS terms spot instances). Sometimes the API doesn't allow booting workers due to unavailable resources and throws an error.

So the following happens:
- API call to create an agent is made and the API call gets an error (resource unavailable)
- an agent entry into the database is done in state error
- in the situation of 1 pending job the autoscaler thinks there are 1 pending jobs and capacity is 1.

If one relies solely on the autoscaler it's a stalemate. the job never gets picked up 😢 


Out of curiosity in your described scenario. The agent never get's fully finished or am I wrong here?
Wouldn't the agent stay in error state, count as available capacity (while it's not available)?

Maybe there is a retry since it's a different scenario than what we are experiencing.

Admittedly drone doesn't handle agent killing gracefully, but the pricing versus the amount of time of it happening, still makes it attractive for us.

bradrydzewski:
currently if an instance is in an error state it requires manual intervention. Repairing or removing servers in an error state is something that will be possible to automate once we know more about the failure points.

genisd:
I'll be building something like this (out of neccessity), but I'd appreciate your input so that we can merge it in the long run.

I was thinking of doing a periodic "zombie" check over all agents to verify that they're operational.

The best check that I can come up with would be to verify docker connectivity and verify that drone-agent is running. If that does not succeed, place the agent in a zombie state which get's garbage collected (say 2 successive failures are needed).

I think that could be done over all agents except those which are booting up/bootstrapping (`if staging && age < 10 min`). This way we should catch all scenarios where there are not fully operational agents which count as capacity.

By doing that I think we'd eliminate all manual intervention needed. We can check and garbage collect error states, half provisioned agents (the cause should not matter and probably differs per cloud provider).

PR #17, opened 2018-05-02
WIP: do a TCP connect on agents and delete agents which are non-responsive

genisd:
this is the WIP code that we're running to spot killed instances.
at the moment it's only a TCP check

still needs tests
also the check should be improved towards a more real test/check that the agent is running

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -62,6 +62,10 @@ type (
 			Datasource string `default:"database.sqlite?cache=shared&mode=rwc&_busy_timeout=9999999"`
 		}
 
+		Zombie struct {
+			MinAge time.Duration `default:"5m"`
+		}
+
 		Amazon struct {
 			Image         string
 			Instance      string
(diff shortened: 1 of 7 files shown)


bradrydzewski:
I actually had a different approach in mind, which pings the docker daemon directly. This verifies not only the server is healthy, but also the docker daemon. The implementation is available in master with the latest image.

The code is here:
https://github.com/drone/autoscaler/blob/master/engine/pinger.go

To enable the pinger you need to set the following environment variable:

```
DRONE_ENABLE_PINGER=true
```

To enable cleanup of error instances you need to set the following environment variable:

```
DRONE_ENABLE_REAPER=true
```

Note that both implementations are experimental and we will eventually remove the feature flags. In the meantime, feel free to post feedback our issues to the [mailing list](https://discourse.drone.io/).

genisd:
@bradrydzewski yes the actual check was indeed very poor but it worked for us.
We're running this right now but if there's a better solution in master we'll definitely make the switch.

From your description it sounds much much better than this implementation.

Thank you!

PR #26, opened 2018-10-16
driver-openstack: Adds OpenStack driver

colin-campbell:

### DIFF
--- a/drivers/openstack/provider.go
+++ b/drivers/openstack/provider.go
@@ -32,7 +32,7 @@ type provider struct {
 }
 
 // New returns a new OpenStack provider.
-func New(opts ...Option) autoscaler.Provider {
+func New(opts ...Option) (autoscaler.Provider, error) {
 	p := new(provider)
 	for _, opt := range opts {
 		opt(p)
@@ -43,21 +43,22 @@ func New(opts ...Option) autoscaler.Provider {
 	}
 
 	if p.computeClient == nil {
-		opts, err := openstack.AuthOptionsFromEnv()
+		authOpts, err := openstack.AuthOptionsFromEnv()
+		if err != nil {
+			return nil, err
+		}
 
+		authClient, err := openstack.AuthenticatedClient(authOpts)
 		if err != nil {
-			panic(err)
+			return nil, err
 		}
-		authClient, err := openstack.AuthenticatedClient(opts)
-		if err != nil {
-			panic(err)
-		}
+
 		p.computeClient, err = openstack.NewComputeV2(authClient, gophercloud.EndpointOpts{
 			Region: p.region,
 		})
 		if err != nil {
-			panic(err)
+			return nil, err
 		}
 	}
-	return p
+	return p, nil
 }
--- a/drivers/openstack/provider.go
+++ b/drivers/openstack/provider.go
@@ -1,0 +1,63 @@
+// Copyright 2018 Drone.IO Inc
+// Use of this software is governed by the Business Source License
+// that can be found in the LICENSE file.
+
+package openstack
+
+import (
+	"sync"
+	"text/template"
+
+	"github.com/drone/autoscaler"
+	"github.com/drone/autoscaler/drivers/internal/userdata"
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack"
+)
+
+// provider implements an OpenStack provider
+type provider struct {
+	init sync.Once
+
+	key      string
+	region   string
+	image    string
+	flavor   string
+	subnet   string
+	pool     string
+	userdata *template.Template
+	groups   []string
+	metadata map[string]string
+
+	computeClient *gophercloud.ServiceClient
+}
+
+// New returns a new OpenStack provider.
+func New(opts ...Option) autoscaler.Provider {
+	p := new(provider)
+	for _, opt := range opts {
+		opt(p)
+	}
+
+	if p.userdata == nil {
+		p.userdata = userdata.T
+	}
+
+	if p.computeClient == nil {
+		opts, err := openstack.AuthOptionsFromEnv()
+
+		if err != nil {
+			panic(err)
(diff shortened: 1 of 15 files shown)


drivers/openstack/provider.go
```diff
@@ -0,0 +1,63 @@
...
+
+	if p.userdata == nil {
+		p.userdata = userdata.T
+	}
+
+	if p.computeClient == nil {
+		opts, err := openstack.AuthOptionsFromEnv()
+
+		if err != nil {
+			panic(err)
```
tboerger: Please don't panic, look at the Google provider, you can gracefully fail if you return an error

drivers/openstack/provider.go
```diff
@@ -0,0 +1,63 @@
...
+	for _, opt := range opts {
+		opt(p)
+	}
+
+	if p.userdata == nil {
+		p.userdata = userdata.T
+	}
+
+	if p.computeClient == nil {
+		opts, err := openstack.AuthOptionsFromEnv()
```
tboerger: You should use a different name here, opts is already an argument of this function.

drivers/openstack/provider.go
colin-campbell: Thanks. I was thinking about that, but while setupProvider() returns an error,  provider.New() does not and the calling function always returns nil.  Is it acceptable to change the signature of openstack.New() to return an error in this case?

colin-campbell: Also. I notice that WithUserDataFile() option panics in all providers.

colin-campbell:
Panics removed and some cleanup + tests added for provider and config.

bradrydzewski:
This is awesome! The only feedback is that I would love to see improved test coverage on the create an destroy operations (following a pattern similar to [this](https://github.com/drone/autoscaler/blob/master/drivers/digitalocean/create_test.go))

colin-campbell:
@bradrydzewski @tboerger I would love that too :) 
But I have spent the entire day attempting to get gock to handle the gophercloud results paging system in OS without any success. A single call to create a server can involve 10+ calls to the OS api listing images, secgroups, flavors, etc, and grok+gophercloud simply does not want to play ball. I shall continue, but it's not looking hopeful. 
I presume this is why the ec2 client lacks these tests too? :)

bradrydzewski:
have you tested to make sure the code works with a live open stack instance?  If yes, we can merge and I can try to help add one or two gock tests in a separate patch. Perhaps if I put some example tests in place, you can help add a few more?

JordanSussman:
This looks awesome!

Some initial feedback/thoughts on the creation step:

It is probably worthwhile to add the logic/option to utilize existing floating IPs instead of [creating](https://github.com/drone/autoscaler/pull/26/files#diff-12af315cedca5b09c056f960fe5c1fa3R30)/deleting them every time. The primary use case for this is related to having firewall rules on FIPs instead of whitelisting the entire OpenStack subnets.

Thoughts?

tboerger:
But he's only creating a floating ip from a specific pool, isn't that enough?

JordanSussman:
> But he's only creating a floating ip from a specific pool, isn't that enough?

Unfortunately, no creating/deleting the floating ip from an external pool isn't enough for our particular use case. To add further context/clarification around some OpenStack architecture:

- Floating IPs can be allocated from a particular pool to a project (tenant)
- Some of the external floating ip pools can be quite large (in our particular case we utilize some /21 and /22 networks)

Ideally, this code could handle the scenario where you give it a list of floating IPs already allocated to the project, instead of always getting random floating IPs from the external pool. Due to various regulatory reasons, we aren't able to blanket the entire external floating IP pool with firewall access. This requires us to allocate a select number of floating IPs from the pool, so we can assign firewall rules to them. 

On a semi-related note, you should at least consider switching from [openstack/compute/v2/extensions/floatingips](https://godoc.org/github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/floatingips) to [openstack/networking/v2/extensions/layer3/floatingips](https://godoc.org/github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips). The particular nova endpoints through `openstack/compute/v2/extensions/floatingips` have been deprecated for quite some time.

colin-campbell:
Sure, I'm open to any ideas, but basically I want to get it merged and a full test suite done, before adding functionality. This is at the "scratching an itch"-stage at the moment. 
The OS API is considerably more tricky to deal with than some other providers, and I want a solid, functional base to work with.

tboerger:
The first goal is to get it generally rolling on OpenStack of course. This should be already archived with some basic test coverage. When this is done we should be able to merge it and to add more flexibility for specific use cases like the one from @JordanSussman.

tboerger:
Maybe you could create feature request issues for these improvements.

JordanSussman:
Doesn't look like https://github.com/drone/autoscaler/issues is available. Should discussion move to https://discourse.drone.io/ instead?

tboerger:
You are right, I forgot that the issues are disabled. Brad should decide where to put such issues.

bradrydzewski:
I went ahead and merged the pull request, that way @JordanSussman can start working on a pull request. If you want to have any design discussion (around the feature) before you start coding, I recommend creating a placeholder pull request, and we can use the github comments to discuss further.

@colin-campbell if you have some time over the weekend (you mentioned previously) to try and get a gock test up and running that would be awesome. Try to pick the simplest function. Otherwise I can take a look early next week and see if I can get an example test in place.

bradrydzewski:
I forgot to mention that the docs are in my private gogs repository. I will move these to GitHub so that we can get Open Stack (and Hetzner) properly documented.

colin-campbell:
@bradrydzewski Tentatively, I think may have discovered  the gock issue. It seems gophercloud is extremely picky about the content-type which has to be explicitly set to application/json for the mocks. I'll open a new PR for the tests

PR #29, opened 2019-02-19
allow host volume mounts for drone agents

patrickjahns:
Today we encountered the issue with drone 1.0 agents, that they can not work with self-signed certificates on the drone server as we are not able to mount our ca.

This PR attempts to provide a environment variable in which one can define arbitrary host to agent-container mounts

### DIFF
--- a/engine/install.go
+++ b/engine/install.go
@@ -7,6 +7,7 @@ package engine
 import (
 	"context"
 	"fmt"
+	"github.com/drone/autoscaler/config"
 	"io"
 	"io/ioutil"
 	"sync"
@@ -31,6 +32,7 @@ type installer struct {
 	proto            string
 	keepaliveTime    time.Duration
 	keepaliveTimeout time.Duration
+	runner 	         config.Runner
 
 	servers autoscaler.ServerStore
 	client  clientFunc
@@ -140,6 +142,9 @@ poller:
 				fmt.Sprintf("DRONE_RPC_SECRET=%s", i.secret),
 				fmt.Sprintf("DRONE_RUNNER_CAPACITY=%v", instance.Capacity),
 				fmt.Sprintf("DRONE_RUNNER_NAME=%s", instance.Name),
+				fmt.Sprintf("DRONE_RUNNER_VOLUMES=%s", i.runner.Volumes),
+				fmt.Sprintf("DRONE_RUNNER_DEVICES=%s", i.runner.Devices),
+				fmt.Sprintf("DRONE_RUNNER_PRIVILEGED_IMAGES=%s", i.runner.Privileged),
 			},
 			Volumes: toVol(i.volumes),
 			Labels: map[string]string{
--- a/engine/install.go
+++ b/engine/install.go
@@ -11,6 +11,7 @@ import (
 	"io/ioutil"
 	"sync"
 	"time"
+	"regexp"
 	"strings"
 
 	"github.com/drone/autoscaler"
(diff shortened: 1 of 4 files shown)


bradrydzewski:
the only comment I have is that there are technically two different volume settings. There are volumes we want to mount for the agent, and there are volumes we want to mount to the pipelines. Thoughts only how to differentiate naming?

engine/install.go
```diff
@@ -126,6 +128,20 @@ poller:
...
+
+	binds := []string{
+		"/var/run/docker.sock:/var/run/docker.sock",
+	}
+
+	for _, volume := range i.volumes {
+		splitted := strings.Split(volume, ":")
+		volumes[splitted[1]] = struct{}{}
+		binds = append(binds, volume)
+	}
```
bradrydzewski: lets re-use this helper function:
```
// helper function that converts a slice of volume paths to a set of
// unique volume names.
func toVol(paths []string) map[string]struct{} {
	set := map[string]struct{}{}
	for _, path := range paths {
		parts, err := splitVolumeParts(path)
		if err != nil {
			continue
		}
		if len(parts) < 2 {
			continue
		}
		set[parts[1]] = struct{}{}
	}
	return set
}
```

and also lets import the unit test:
https://github.com/cncd/pipeline/blob/master/pipeline/backend/docker/convert_test.go#L78

tboerger:
> the only comment I have is that there are technically two different volume settings. There are volumes we want to mount for the agent, and there are volumes we want to mount to the pipelines. Thoughts only how to differentiate naming?

There are already different names... You named the pipeline volumes DRONE_RUNNER_VOLUMES, that's why I suggested DRONE_AGENT_VOLUMES.

engine/install.go
patrickjahns: @bradrydzewski 
done - however I am a bit concerned with the authorship - code is partially from you and partially from another contributor

patrickjahns:
I can also expose a `DRONE_RUNNER_VOLUMES` in a different PR

patrickjahns:
Drone failure is some github issue:
```
+ git init
Initialized empty Git repository in /go/src/github.com/drone/autoscaler/.git/
+ git remote add origin https://github.com/drone/autoscaler.git
+ git fetch --no-tags origin +refs/pull/29/merge:
error: RPC failed; curl 18 transfer closed with outstanding read data remaining
fatal: The remote end hung up unexpectedly
exit status 128
```

engine/install.go
```diff
@@ -126,6 +131,12 @@ poller:
 		Str("image", i.image).
 		Msg("create agent container")
 
+	volumes := toVol(i.volumes)
```
tboerger:
```
i.volumes = append(i.volumes, "/var/run/docker.sock:/var/run/docker.sock")

<SNIP>

Volumes: toVol(i.volumes),

<SNIP>

Binds: i.volumes,
```

I would suggest something like that.

engine/install.go
```diff
@@ -126,6 +131,7 @@ poller:
 		Str("image", i.image).
 		Msg("create agent container")
 
+	i.volumes = append(i.volumes, "/var/run/docker.sock:/var/run/docker.sock")
```
bradrydzewski: this is not thread safe. Instead of mutating `i.volumes` a local variable should be used:

```
var volumes = append(i.volumes, "/var/run/.....")
```

patrickjahns: Thanks for pointing this out - just pushed a fix

bradrydzewski:
great, thanks! I think this looks good for now. I will merge once the PR passes test.

PR #31, opened 2019-02-21
Fixed issue where the DRONE_DIGITALOCEAN_IMAGE environment variable would not be used.

jlesage:
As reported here: https://discourse.drone.io/t/autoscaler-is-using-an-invalid-digital-ocean-image/3832, this fix the issue where `DRONE_DIGITALOCEAN_IMAGE` environment variable is ignored.


bradrydzewski:
good catch!

PR #32, opened 2019-02-21
Apply standard format, add multi arch docker builds

tboerger:


.gitignore
```diff
@@ -1,9 +1,31 @@
-NOTES.md
-release
-vendor
-*.sqlite
-*.sqlite3
-*.bak
-*.out
-*.db
-*.env
```
bradrydzewski: any reason we need to change the .gitignore?  the newer version specifies a lot of files that will never exist in the project, but excludes files that should be ignored like sqlite files.

.drone.yml
```diff
@@ -1,78 +1,469 @@
...
-  path: src/github.com/drone/autoscaler
-
 steps:
+- name: vet
+  pull: always
+  image: golang:1.11
+  commands:
+  - go vet ./...
+  environment:
+    GO111MODULE: on
```
bradrydzewski: remove `GO111MODULE: on` since it is not necessary when building outside the gopath, and remove `pull: always` because it prevents me from using `drone exec` locally when I don't have internet (like when I'm working from a coffee shop). This same feedback should apply to all official Drone repos in the `drone/` org. You can use slightly different standard for `drone-plugins/` as you see fit.

.drone.yml
```diff
@@ -1,78 +1,469 @@
...
+      from_secret: github_token
+  when:
+    event:
+    - tag
+
+trigger:
+  ref:
+  - refs/heads/master
+  - "refs/tags/**"
+  - "refs/pull/**"
```
bradrydzewski: these should be removed because it prevents me from testing feature branches.  I understand this is a standard for drone-plugins, but is not a standard I want to impose in `drone/` org.

bradrydzewski:
hey, I am reviewing this PR and finding some problems. Won't these changes break the published docker image?  This project uses sqlite and needs special ldflags that do not exist in the libsonnet file that will cause a runtime error? There are four separate types of projects in the drone org, each of which requires a different pipeline structure:

1. docker projects with static compilation
2. docker projects without static compilation
3. library projects
4. CLI projects

tboerger:
You are right, I will revisit my open prs to the drone org and make it more simple.

.gitignore
tboerger: Yeah the sqlite extensions are indeed missing now... The rest is just a copy from some standard go gitignore file.

.drone.yml
tboerger: Will fix that for all the prs

.drone.yml
tboerger: I will drop that expect for the release pipeline steps

docker/Dockerfile.linux.amd64
```diff
@@ -1,16 +1,18 @@
-FROM alpine:3.6 as alpine
-RUN apk add -U --no-cache ca-certificates
+FROM alpine:3.9
+
+LABEL maintainer="Drone.IO Community <drone-dev@googlegroups.com>" \
+  org.label-schema.name="Drone Autoscaler" \
+  org.label-schema.vendor="Drone.IO Community" \
+  org.label-schema.schema-version="1.0"
 
-FROM alpine:3.6
```
jones2026: @tboerger, any reason to not use `FROM scratch` here and just import the certs from the alpine image?

jones2026:
Thank you so much for converting to Go mod instead of go dep!

PR #33, opened 2019-05-08
Ignoring the capacity of errored servers

andy-trimble:
In the case where a server cannot be created (due to an AWS issue, for example), the non-existent server's capacity is still used to calculate total capacity. This seems like undesirable behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+### Added
+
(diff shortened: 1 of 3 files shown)


andy-trimble:
Ah, that makes sense. I ran into an issue where the EC2 instance could not be created (due to my misconfiguring the security group), but the autoscaler would not allocate additional resources since it believed the capacity had been filled. I wonder if there is a way to detect when the instance has not been created and ignore the capacity in that instance.

tboerger:
For some cases I wish the autoscaler would exclude the errored nodes, at least for some providers it would be possible to check if there is really some server living.

bradrydzewski:
we have a process that tries to remove errored instances:
https://github.com/drone/autoscaler/blob/master/engine/reaper.go

tboerger:
> We have a process that tries to automatically remove errored instances:
> https://github.com/drone/autoscaler/blob/master/engine/reaper.go

Good to know that it's currently not working for Hetzner, got to fix that.

PR #41, opened 2019-07-18
agent label assignment and matching

logikone:
* allow matching against stages with specific tags
* agent/planner labels are set via the `DRONE_AGENT_LABELS` environment
variable


bradrydzewski:
Thanks Chris, this looks really good. I should have some time tomorrow morning to review in greater detail and get it merged.

logikone:
right on, ty sir!

logikone:
i noticed a typo in my commit message so i amended it. `s/LABLES/LABELS`

config/config.go
```diff
@@ -49,6 +49,7 @@ type (
 			Kernel      string
 			Environ     []string
 			Volumes     []string
+			Labels      []string `envconfig:"DRONE_AGENT_LABELS"`
```
msaizar: I've been working on something similar, used 
```go
Labels      map[string]string `envconfig:"DRONE_AGENT_LABELS"`
```

instead, `envconfig` will handle `DRONE_AGENT_LABELS=mem:high,cpu:load`.

engine/install.go
```diff
@@ -36,6 +36,7 @@ type installer struct {
 	keepaliveTime    time.Duration
 	keepaliveTimeout time.Duration
 	runner           config.Runner
+	labels 			 []string
```
msaizar: Same here with `map[string]string`

engine/planner.go
```diff
@@ -29,6 +30,7 @@ type planner struct {
 	max     int           // max number of servers to allocate
 	cap     int           // capacity per-server
 	ttu     time.Duration // minimum server age
+	labels  []string
```
msaizar: And here

engine/planner.go
```diff
@@ -268,8 +270,29 @@ func (p *planner) listBusy(ctx context.Context) (map[string]struct{}, error) {
...
+
 	return stage.OS == p.os &&
 		stage.Arch == p.arch &&
 		stage.Variant == p.version &&
-		stage.Kernel == p.kernel
+		stage.Kernel == p.kernel &&
+		labelMatch
+}
+
+func (p *planner) matchLabel(stage *drone.Stage) bool {
```
msaizar: This wouldn't be necessary

engine/planner.go
```diff
@@ -268,8 +270,29 @@ func (p *planner) listBusy(ctx context.Context) (map[string]struct{}, error) {
...
+	if len(p.labels) > 0 || len(stage.Labels) > 0 {
+		labelMatch = p.matchLabel(stage)
+	}
+
 	return stage.OS == p.os &&
 		stage.Arch == p.arch &&
 		stage.Variant == p.version &&
-		stage.Kernel == p.kernel
+		stage.Kernel == p.kernel &&
+		labelMatch
```
msaizar: We could match with `reflect.DeepEqual(stage.Labels, p.labels)` instead

engine/install.go
```diff
@@ -158,6 +159,12 @@ poller:
 		fmt.Sprintf("DRONE_RUNNER_PRIVILEGED_IMAGES=%s", i.runner.Privileged),
 	)
 
+	if len(i.labels) > 0 {
+		envs = append(envs,
+			fmt.Sprintf("DRONE_RUNNER_LABELS=%s", strings.Join(i.labels, ",")),
```
msaizar: Not 100% sure here, I think with the `map` we could just do `fmt.Sprintf("DRONE_RUNNER_LABELS=%s", i.labels),` in the big `append` above

config/config.go
bradrydzewski: this is true.  we did not use maps for environment variables (not sure why) but we did use maps for other configuration parameters. I think in this case using a map makes sense.

engine/planner.go
```diff
@@ -268,8 +270,29 @@ func (p *planner) listBusy(ctx context.Context) (map[string]struct{}, error) {
...
+}
+
+func (p *planner) matchLabel(stage *drone.Stage) bool {
+	plannerLabels := formatLabels(p.labels)
+
+	for key, val := range stage.Labels {
+		if plannerVal, ok := plannerLabels[key]; ok {
+			if strings.EqualFold(plannerVal, val) {
+				return true
```
bradrydzewski: this should return false if a single label does not match, and should return true only if all labels match.

engine/planner.go
bradrydzewski: this probably comes down to personal preference, but I prefer the current approach of using a typed function.

engine/install.go
bradrydzewski: if you have `map[string]string{"foo":"bar"}` it would print `map[foo:bar]`

msaizar: Thanks, I hadn't tested this bit, completely new to `go`. Sorry for the bad advice!

config/config.go
logikone: nice, i didn't realize `envconfig` handled maps. i'll get these all updated

engine/planner.go
logikone: ok cool, ya i wasn't sure on that one. i'll use `reflect.DeepEqual` like @msaizar mentioned above

bradrydzewski: I recommend copying the exact same function we are using to match on the server, to ensure the logic syncs up

https://github.com/drone/drone/blob/master/scheduler/queue/queue.go#L232
https://github.com/drone/drone/blob/master/scheduler/queue/queue.go#L175

logikone: ah, ok. will do

logikone: ok, i've got that updated

engine/formatlabels.go
```diff
@@ -0,0 +1,19 @@
+package engine
+
+import (
+	"strings"
+)
+
+// formatLabels takes a comma separated list of agent labels
+// and formats it into map[string]string for agent label matching
+func formatLabels(labels []string) map[string]string {
```
bradrydzewski: is this still being used?

bradrydzewski:
thanks Chris for the changes. Sorry to pull you onto your computer on a Saturday!

engine/formatlabels.go
logikone: no, i was just realizing that myself. i'll pull this out right now

logikone:
no worries, thank you for the input!

PR #42, opened 2019-08-01
Failed to ping server fix

msaizar:
The server will still be in a running state after ping failure, and the autoscaler will keep trying to reprovision it. The stopped time is set, but the state is not updated. This puts it in a stopped state, similar to https://github.com/drone/autoscaler/blob/master/engine/collect.go#L108-L109 and https://github.com/drone/autoscaler/blob/master/engine/reaper.go#L121-L122

### DIFF
--- a/engine/pinger.go
+++ b/engine/pinger.go
@@ -98,6 +98,6 @@ func (p *pinger) ping(ctx context.Context, server *autoscaler.Server) error {
 
 	server.Error = "Failed to ping the server"
 	server.Stopped = time.Now().Unix()
-	server.State = autoscaler.StateStopped
+	server.State = autoscaler.StateError
 	return p.servers.Update(ctx, server)
 }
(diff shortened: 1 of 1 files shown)


engine/pinger.go
```diff
@@ -98,5 +98,6 @@ func (p *pinger) ping(ctx context.Context, server *autoscaler.Server) error {
 
 	server.Error = "Failed to ping the server"
 	server.Stopped = time.Now().Unix()
+	server.State = autoscaler.StateStopped
```
bradrydzewski: I think this should be `autoscaler.StateError` to indicate the server is in an unexpected state. You can use a feature flag to enable Drone to garbage collect errored servers.

bradrydzewski:
good catch. I posted a minor comment with regards to which status I think we should use, but otherwise should be good to merge.

msaizar:
Updated. The reaper took care of it eventually I think but I'll look into the GC, I haven't tried it yet, thank you! I also noticed that the pinger is not closing connections. On long running agents, you can see a lot of docker connections from the autoscaler with `netstat -an`, one for every 10 minutes the agent was up. Haven't figured out a fix yet but I'll open a discourse topic on that

bradrydzewski:
thanks!

>  I also noticed that the pinger is not closing connections

the `Ping` function is actually the Docker Client code. So if there is a connection leak it would unfortunately be in the Docker client library. So we would need to send the a patch.

PR #43, opened 2019-08-02
Replaced APIClient with Client, added Close()s

msaizar:
As mentioned [here](https://github.com/drone/autoscaler/pull/42#issuecomment-517486281,) I've seen connection issues with the autoscaler. `netstat -an | grep 2376` will show multiple `ESTABLISHED` connections from the autoscaler after some time. 

Digging through the docker-go repo I ran into [this](https://github.com/docker/go-docker/issues/16) and [this](https://github.com/golang/gddo/issues/623). I've applied a similar solution to what he did. `APIClient` doesn't seem to implement `Close`, so I've replaced it with `Client`, and added `defer client.Close()` whenever one is created. Pinger seems to work as before, and no more dangling `ESTABLISHED` connections are seen.

### DIFF
--- a/engine/collect.go
+++ b/engine/collect.go
@@ -84,6 +84,7 @@ func (c *collector) collect(ctx context.Context, server *autoscaler.Server) erro
 	if err != nil {
 		return err
 	}
+	
 	timeout := time.Hour * 60
 	err = client.ContainerStop(ctx, "agent", &timeout)
 	if err != nil {
--- a/engine/collect.go
+++ b/engine/collect.go
@@ -80,10 +80,10 @@ func (c *collector) collect(ctx context.Context, server *autoscaler.Server) erro
 	}
 
 	client, err := c.client(server)
+	defer client.Close()
 	if err != nil {
 		return err
 	}
-
 	timeout := time.Hour * 60
 	err = client.ContainerStop(ctx, "agent", &timeout)
 	if err != nil {
--- a/engine/docker.go
+++ b/engine/docker.go
@@ -18,11 +18,11 @@ import (
 // clientFunc defines a builder funciton used to build and return
 // the docker client from a Server. This is primarily used for
 // mock unit testing.
-type clientFunc func(*autoscaler.Server) (docker.APIClient, error)
+type clientFunc func(*autoscaler.Server) (*docker.Client, error)
 
 // newDockerClient returns a new Docker client configured for the
 // Server host and certificate chain.
-func newDockerClient(server *autoscaler.Server) (docker.APIClient, error) {
+func newDockerClient(server *autoscaler.Server) (*docker.Client, error) {
 	tlsCert, err := tls.X509KeyPair(server.TLSCert, server.TLSKey)
 	if err != nil {
 		return nil, err
--- a/engine/install.go
+++ b/engine/install.go
@@ -90,6 +90,7 @@ func (i *installer) install(ctx context.Context, instance *autoscaler.Server) er
 		Logger()
 
 	client, err := i.client(instance)
+	defer client.Close()
 	if err != nil {
 		logger.Error().Err(err).
 			Msg("cannot create docker client")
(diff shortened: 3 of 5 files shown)


bradrydzewski:
awesome.  do you think we could do something like this instead?

```diff
-func(*autoscaler.Server) (docker.APIClient, error)
+func(*autoscaler.Server) (docker.APIClient, io.Closer, error)
```

This would allow us to keep interfaces which is helpful for the unit tests since they can be mocked. The engine package does not have as much coverage as the rest of the codebase, but that is something I would like to improve over time and interfaces should help.

Thanks again!

msaizar:
I'm not 100% sure how to do this, new to go, but I'll investigate and try things around

engine/collect.go
```diff
@@ -79,7 +79,8 @@ func (c *collector) collect(ctx context.Context, server *autoscaler.Server) erro
 		Size:     server.Size,
 	}
 
-	client, err := c.client(server)
+	client, closer, err := c.client(server)
+	defer closer.Close()
```
bradrydzewski: this is subject to a nil pointer exception, so you'll want the following:

```
if closer != nil {
  defer closer.Close()
}
```

(same feedback for other parts of the code that invokes the closer)

engine/docker.go
```diff
@@ -39,5 +40,6 @@ func newDockerClient(server *autoscaler.Server) (docker.APIClient, error) {
 		},
 	}
 	host := fmt.Sprintf("https://%s:2376", server.Address)
-	return docker.NewClient(host, api.DefaultVersion, client, nil)
+	new_client, err := docker.NewClient(host, api.DefaultVersion, client, nil)
```
bradrydzewski: Go style guide prefers camel case, so you could call this something like `dockerClient`, to differentiate from the http client defined above.

engine/install.go
```diff
@@ -89,7 +89,8 @@ func (i *installer) install(ctx context.Context, instance *autoscaler.Server) er
 		Str("name", instance.Name).
 		Logger()
 
-	client, err := i.client(instance)
+	client, closer, err := i.client(instance)
+	defer closer.Close()
```
bradrydzewski: `if closer != nil`

engine/pinger.go
```diff
@@ -59,7 +59,8 @@ func (p *pinger) ping(ctx context.Context, server *autoscaler.Server) error {
 		Str("name", server.Name).
 		Logger()
 
-	client, err := p.client(server)
+	client, closer, err := p.client(server)
+	defer closer.Close()
```
bradrydzewski: `if closer != nil`

msaizar:
Thanks for the handholding, and on a weekend too

bradrydzewski:
Thanks for making the changes. I am happy to help. Before we can merge it looks like the build is failing due to compiler errors:

```
FAIL	github.com/drone/autoscaler/engine [build failed]
```

msaizar:
From the errors

```
engine/collect_test.go:44:4: cannot use client (type *mocks.MockAPIClient) as type io.Closer in return argument:
*mocks.MockAPIClient does not implement io.Closer (missing Close method)
```

https://github.com/drone/autoscaler/pull/43/files#diff-13241343e97d5a25876915b85e86c039R43-R44

Tried something like this in `mocks/mock_docker.go` but no luck

```go
// Close mocks base method
func (m *MockAPIClient) Close() error {
        ret := m.ctrl.Call(m, "Close")
        ret0, _ := ret[0].(error)
        return ret0
}

// Close indicates an expected call of Close
func (mr *MockAPIClientMockRecorder) Close() *gomock.Call {
        return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "Close", reflect.TypeOf((*MockAPIClient)(nil).Close))
}
```

bradrydzewski:
for the unit tests you can return `nil` for the `io.Closer`. for example:

```diff
	c := collector{
		servers:  store,
		provider: provider,
		client: func(*autoscaler.Server) (docker.APIClient, io.Closer, error) {
-			return client, client, nil
+			return client, nil, nil
		},
	}
```

since the closer is only invoked when non-nil this should ensure the unit tests pass.

msaizar:
That did it! Thanks again

PR #45, opened 2019-08-24
Implement scaleway driver

frebib:
This is WIP, but works in it's current state. Tests to follow.

Scaleway is one of the only providers that support ARM architecture VMs which we are in need of, so we decided this was the right approach

### DIFF
--- a/cmd/drone-autoscaler/main.go
+++ b/cmd/drone-autoscaler/main.go
@@ -19,6 +19,7 @@ import (
 	"github.com/drone/autoscaler/drivers/hetznercloud"
 	"github.com/drone/autoscaler/drivers/openstack"
 	"github.com/drone/autoscaler/drivers/packet"
+	"github.com/drone/autoscaler/drivers/scaleway"
 	"github.com/drone/autoscaler/engine"
 	"github.com/drone/autoscaler/metrics"
 	"github.com/drone/autoscaler/server"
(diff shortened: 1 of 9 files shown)


frebib:
CI is failing because I can't even `go dep`. I've never used it before and I can't get it to work. Some pointers would be appreciated, please

frebib:
I have reworked the way that this destroys old servers, hence the repeated changes, but I think the logic is simple and stable now

frebib:
> You should not panic at all.

I agree. I was being consistent with other drivers in this case. Given the function signature doesn't handle errors, what else could/should I do?

tboerger:
IMHO you should check that within the New func and error out.

frebib:
Option errors are checked and handled gracefully now

tboerger:
Looks like your choosen default values aren't that great, the autoscaler always fails for C1 and ARM64-2GB instances, local boot mode is not possible.

frebib:
The image/size combo is a bit complicated. As default it _should_ work with amd64. You have to change the image if you change the arch, location or size

tboerger:
Looks like I got to create a PR to fix the options, https://github.com/drone/autoscaler/blob/master/drivers/scaleway/create.go#L35 this hard-coded value doesn't work for the arm32/arm64 images. I don't need scaleway for amd64 nodes, I need it only for arm32/arm64.

ashwilliams1:
while we're on the topic, it would be great to have some docs for Scaleway at https://github.com/drone/autoscaler-docs 😄

tboerger:
https://github.com/drone/autoscaler/pull/55 will fix my issues with the Scaleway implementation, after that I can prepare a PR that documents this provider.

frebib:
I started a branch for documenting this, but I haven't had time to finish it. You're welcome to use it as a base
https://github.com/frebib/autoscaler-docs/commit/858d6bc41d38e6da57d1a79563ec9e878aded6cf

Blafy:
Hi !
@frebib Nice feature to have the Scaleway implementation, thanks.
But the docs aren't merged yet.
What is missing in your documentation ? So we can complete it and PR it ?

PR #46, opened 2019-08-24
Use gomod instead of godep

frebib:
I think the title is explanatory enough. 

Simply `go mod init <repo>` plus some cleanup


tboerger:
Go mod is already part of https://github.com/drone/autoscaler/pull/32

frebib:
#32 hasn't had movement in months though

bradrydzewski:
this should mount a temp volume to share the go module cache between pipeline steps (except publish), otherwise each step will re-download the deps. See https://docs.drone.io/examples/language/golang/#dependencies

PR #48, opened 2019-09-03
Ignore error state on planner

msaizar:
I'm not 100% sure if this would be the best way to go. 

Currently an instance will be placed in an error state if there is no spot capacity available in AWS for example. These will count towards the autoscaler's capacity, and won't be destroyed until the reaper gets rid of them in the next hour. I made a small change so that instances in error state won't count just like stopped instances. 

Another option would be to move the reaper interval to an environment variable, so that it can be run every 10 minutes for example. While it is true that this is infrequent, clearing these every hour could mean a queue building for an hour until they're cleared.

### DIFF
--- a/engine/planner.go
+++ b/engine/planner.go
@@ -242,6 +242,8 @@ func (p *planner) capacity(ctx context.Context) (capacity, count int, err error)
 		switch server.State {
 		case autoscaler.StateStopped:
 			// ignore state
+		case autoscaler.StateError:
+			// ignore state
 		default:
 			count++
 			capacity += server.Capacity


bradrydzewski:
> Currently an instance will be placed in an error state if there is no spot capacity available in AWS for example. These will count towards the autoscaler's capacity, and won't be destroyed until the reaper gets rid of them in the next hour. 

this is by design.  See my comments in https://github.com/drone/autoscaler/pull/33, which was a similar proposal that we did not merge.

> Another option would be to move the reaper interval to an environment variable, so that it can be run every 10 minutes for example. While it is true that this is infrequent, clearing these every hour could mean a queue building for an hour until they're cleared.

the ability to configure the reaper would be a more appropriate change.

msaizar:
Thanks for the quick reply! Should I close this and re open a new PR when I have the code for this, or work on this same PR as a WIP?

bradrydzewski:
lets close and do a new pull request

PR #49, opened 2019-09-03
Reaper interval environment variable

msaizar:
Currently an instance will be placed in an error state if there is no spot capacity available in AWS for example. These will count towards the autoscaler's capacity, and won't be destroyed until the reaper gets rid of them in the next hour. I moved the reaper interval to an environment variable so that one can choose the frequency.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -66,7 +66,6 @@ type (
 		}
 
 		Reaper struct {
-			Enabled  bool          `envconfig:"DRONE_ENABLE_REAPER" default:"false"`
 			Interval time.Duration `envconfig:"DRONE_REAPER_INTERVAL" default:"1h"`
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -65,6 +65,11 @@ type (
 			Cache    string        `envconfig:"DRONE_GC_CACHE" default:"10gb"`
 		}
 
+		Reaper struct {
+			Enabled  bool          `envconfig:"DRONE_ENABLE_REAPER" default:"false"`
```
tboerger: Enabled isn't used yet, right? It's currently based directly on the env variable without using this config.

msaizar: You're right, I've removed it for now

PR #51, opened 2019-09-15
ignore state shutdown/stopping/error from available capacity. cleanup failed server agent bootstrap attempts and retry

genisd:
on install failure dont set to Error, but set to Shutdown so it gets deleted

consider two possible failures
- ppa offline -> after node boot docker cannot be installed
- dockerhub docker images repo offline -> agent cannot be provisioned

situation:
1 pending job, 1 agent needs to be bootstrapped

Old behaviour ( old === 2 months+ ago ).

- server gets booted.
- agent cannot get fully provisioned such that it can handle the pending job
- agent goes into error state
- autoscaler ends up in starvation. job remains forever pending. since error state counts towards available capacity
- booted agent (server) remains online until intervention costing money, job stays pending

New behaviour (this PR)
- server gets booted.
- agent cannot get fully provisioned such that it can handle the pending job
- agent goes into shutdown state, and gets removed from cloud provider
- above is repeated until the error condition disappears and a fully working agent can be provisioned which will handle the job

TLDR:
old situation: starvation
new situation: retry infinitely

The above described error conditions are not uncommon. I've seen both and its not automatically handled in master (again as of 2 month ago, but the changes which were merged will not solve the starvation issue)


engine/install.go
```diff
@@ -358,7 +358,7 @@ func (i *installer) setupGarbageCollector(ctx context.Context, client docker.API
 
 func (i *installer) errorUpdate(ctx context.Context, server *autoscaler.Server, err error) error {
```
genisd: this fn name maybe somewhat incorrect now in regards to this PR

bradrydzewski:
> new situation: retry infinitely

while I appreciate the pull request, infinite retries are the exact thing we are trying to avoid.  Instead, I recommend enabling the reaper to cleanup errored instances, combined with https://github.com/drone/autoscaler/pull/49 to run the reaper more frequently.

genisd:
I figured you might close it, no worries. I do respecfully disagree that
starvation before reaper acts is a better solution, just curious how long
does it take?

In particular because the reaper is disabled by default ( we use it happily
though thanks! ).

Since docker ppa and dockerhub image hosting are a bit flaky at times, new
users using this will get the starvation out of the box :/ And not seldomly
im afraid.

I made the PR because someone on discord asked for this exact behaviour. I
expected thus behaviour myself when starting to use the autoscaler.

The argument that leaving errored nodes behind, paying for it and counting
them towards available capacity, is intended default behaviour i cannot
understand. Having starvation and paying for it does not seem advantageous
to me. Instead retrying and trying to provision an agent to be able to
handle the pending job must be the right way to go.
Nodes incorrecly provisioned are immediately cleaned up.

But oh well, my 0.02c i guess :)

genisd:
And yes you're right https://github.com/drone/autoscaler/pull/49 is a very very good improvement.
It should be enabled by default, and we'll be sure to use set/reduce that timeout in our environment :-)

bradrydzewski:
> The argument that leaving errored nodes behind, paying for it and counting
them towards available capacity, is intended default behaviour i cannot
understand

because we want to prevent the potential for infinite loops of creating and destroying nodes, or situations where the system continues creating nodes but not properly destroying them, which could run up server bills. The consequence for doing so could be a lawsuit which I prefer to avoid. So we take a conservative approach by default, and we provide settings that allow teams to take more risk once they are more comfortable running the autoscaler.

genisd:
I understand where you're coming from.

I think this situation (failed node boot / agent bootstrap) should resolve/cleanup itself by default. I think that is currently not the case which is a bummer for new users.

Maybe it's worth considering chaning the reaper from opt-in to opt-out.

We scale down to 0 nodes, together with low time to live (few minutes). These issues happen on a weekly basis usually, hence why I think it should fix itself by default for best new user experience and adoption.

PR #53, opened 2019-09-21
docker login opt

parlir:
Adds support for configuration option to use a private docker image.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -57,6 +57,8 @@ type (
 			Environ     []string
 			Volumes     []string
 			Labels      map[string]string `envconfig:"DRONE_AGENT_LABELS"`
+			Dockerusername string         `envconfig:"DRONE_AGENT_DOCKER_USERNAME"`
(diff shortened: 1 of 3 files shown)


tboerger:
Why do you need to authenticate at this stage?

engine/install.go
```diff
@@ -143,9 +147,23 @@ poller:
 
 	logger.Debug().
 		Str("image", i.image).
-		Msg("pull docker image")
+			Msg("pull docker image")
```
frebib: Whitespace error here

config/config.go
```diff
@@ -57,6 +57,8 @@ type (
 			Environ     []string
 			Volumes     []string
 			Labels      map[string]string `envconfig:"DRONE_AGENT_LABELS"`
+			Dockerusername string         `envconfig:"DRONE_AGENT_DOCKER_USERNAME"`
```
frebib: Indentation. You should run `gofmt -w **/*.go`

kmk142789 (approved):

PR #54, opened 2019-10-02
Azure support

parlir:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ vendor
 *.out
 *.db
 *.env
+*#
+*.#*
+*~
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -108,3 +108,11 @@
 [[constraint]]
   name = "github.com/lib/pq"
   version = "1.0.0"
+
+[[constraint]]
+  name = "github.com/Azure/azure-sdk-for-go"
+  version = "32.5.0"
+
+[[override]]
+  name = "github.com/Azure/go-autorest"
+  version = "13.0.0"
--- a/drivers/azure/create.go
+++ b/drivers/azure/create.go
@@ -3,3 +3,116 @@
 // that can be found in the LICENSE file.
 
 package azure
+
+import (
+	"context"
+	"github.com/drone/autoscaler"
+	"github.com/rs/zerolog/log"
+	"fmt"
+	"time"
+
+	"bytes"
+	"encoding/base64"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
+	"github.com/Azure/go-autorest/autorest/to"
+)
+
+func (p *provider) computeMachineRequest(ctx context.Context, base string, opts autoscaler.InstanceCreateOpts) (compute.VirtualMachine, *string, error) {
+	buf := new(bytes.Buffer)
+	err := p.userdata.Execute(buf, &opts)
+	cloudInit := base64.StdEncoding.EncodeToString(buf.Bytes())
+
+	nic, ip, err := p.CreateNIC(ctx, p.vnet, p.subnet, p.nsg, base+"-ip", base+"-nic")
+	if err != nil {
+		return compute.VirtualMachine{}, nil, err
+	}
+
+	return compute.VirtualMachine{
+		Location: to.StringPtr(p.location),
+		VirtualMachineProperties: &compute.VirtualMachineProperties{
+			HardwareProfile: &compute.HardwareProfile{
+				VMSize: p.vmSize,
+			},
+			StorageProfile: &compute.StorageProfile{
+				OsDisk: &compute.OSDisk{
+					Name:         to.StringPtr(base + "-" + "osdisk"),
+					CreateOption: compute.DiskCreateOptionTypesFromImage,
+					DiskSizeGB:   to.Int32Ptr(p.volumeSize),
+				},
+				ImageReference: &compute.ImageReference{
+					Offer:     to.StringPtr(p.imageOffer),
+					Publisher: to.StringPtr(p.imagePublisher),
+					Sku:       to.StringPtr(p.imageSKU),
+					Version:   to.StringPtr(p.imageVersion),
+				},
+			},
+			NetworkProfile: &compute.NetworkProfile{
+				NetworkInterfaces: &[]compute.NetworkInterfaceReference{
+					{
+						ID: nic.ID,
+						NetworkInterfaceReferenceProperties: &compute.NetworkInterfaceReferenceProperties{
+							Primary: to.BoolPtr(true),
+						},
+					},
+				},
+			},
+			OsProfile: &compute.OSProfile{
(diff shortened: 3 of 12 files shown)


drivers/azure/create.go
```diff
@@ -3,3 +3,116 @@
...
+
+	azureInstance, err := future.Result(client)
+	client.Start(ctx, p.resourceGroup, p.vmName)
+
+	if err != nil {
+		return nil, err
+	}
+
+	fmt.Printf("Starting sleep: %v\n", time.Now().Unix())
+	time.Sleep(180 * time.Second)
```
parlir: I should probably alter this to just poll on the static IP I provide until it's available. This was a quick hack to fix an issue I had noticed around an agent getting assigned at times +1 builds.

drivers/azure/destroy.go
```diff
@@ -3,3 +3,209 @@
...
+
+	err = future.WaitForCompletionRef(ctx, diskClient.Client)
+	logger.Debug().Msg("waitforcompletionref")
+	if err != nil {
+		logger.Error().
+			Err(err).
+			Msg("cannot terminate instance")
+		return err
+	}
+	return nil
```
bradrydzewski: I noticed lines 44-51 and 60-67 look to be the same

drivers/azure/destroy.go
```diff
@@ -3,3 +3,209 @@
...
+			Msg("cannot terminate instance")
+		return err
+	}
+	
+	if err != nil {
+		logger.Error().
+			Err(err).
+			Msg("cannot terminate instance")
+		return err
+	}
```
bradrydzewski: this error is already handled at line 46

drivers/azure/nic.go
```diff
@@ -0,0 +1,132 @@
...
+	auth, err := auth.NewAuthorizerFromEnvironment()
+	if err != nil {
+		return ipClient, err
+	}
+	ipClient.Authorizer = auth
+	return ipClient, nil
+}
+
+// CreatePublicIP creates a new public IP
+func (p *provider) createPublicIP(ctx context.Context, ipName string) (ip network.PublicIPAddress, err error) {
```
bradrydzewski: is creating a static public IP necessary?  can we just use the ephemeral public IP address associated with the instance? This is what we do for other cloud providers.

engine/install.go
```diff
@@ -40,6 +42,8 @@ type installer struct {
 	keepaliveTimeout time.Duration
 	runner           config.Runner
 	labels           map[string]string
+	dockerUsername         string
+	dockerPassword         string
```
bradrydzewski: these changes can be removed. We have a separate PR that is proposing this change

Gopkg.toml
```diff
@@ -108,3 +108,11 @@
...
   name = "github.com/lib/pq"
   version = "1.0.0"
+
+[[constraint]]
+  name = "github.com/Azure/azure-sdk-for-go"
+  version = "32.5.0"
+
+[[override]]
+  name = "github.com/Azure/go-autorest"
+  version = "13.0.0"
```
bradrydzewski: the `Gopkg` files were removed and replaced with `go.mod` files. You may want to sync up your codebase with the latest version

drivers/azure/create.go
bradrydzewski: yes this should use polling, which is how we handle this for the other providers. You can use this example as reference https://github.com/drone/autoscaler/blob/master/drivers/amazon/create.go#L122:L175

drivers/azure/create.go
```diff
@@ -3,3 +3,116 @@
 // that can be found in the LICENSE file.
 
 package azure
+
+import (
+	"context"
+	"github.com/drone/autoscaler"
+	"github.com/rs/zerolog/log"
+	"fmt"
+	"time"
```
bradrydzewski: best practice for import ordering  https://github.com/golang/go/wiki/CodeReviewComments#imports

drivers/azure/create.go
```diff
@@ -3,3 +3,116 @@
...
+	"bytes"
+	"encoding/base64"
+
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
+	"github.com/Azure/go-autorest/autorest/to"
+)
+
+func (p *provider) computeMachineRequest(ctx context.Context, base string, opts autoscaler.InstanceCreateOpts) (compute.VirtualMachine, *string, error) {
+	buf := new(bytes.Buffer)
+	err := p.userdata.Execute(buf, &opts)
```
bradrydzewski: unhandled exception

bradrydzewski:
thanks for submitting the pull request.  I have added some comments about specific sections of code. I also recommend taking a look at AWS or GCP and trying to more closely mirror the structure and conventions (logging, polling, functions, etc). For example, you will notice we have very detailed logging in the other providers.

.gitignore
```diff
@@ -6,4 +6,7 @@ vendor
 *.bak
 *.out
 *.db
-*.env
\ No newline at end of file
+*.env
+*#
```
bradrydzewski: revert

.gitignore
```diff
@@ -6,4 +6,7 @@ vendor
...
+*.#*
+*~
```
bradrydzewski: what are these lines for?

drivers/azure/create.go
```diff
@@ -3,3 +3,116 @@
...
+		base+"-vm",
+		computeRequest,
+	)
+	
+	err = future.WaitForCompletionRef(ctx, client.Client)
+	if err != nil {
+		return nil, err
+	}
+
+	azureInstance, err := future.Result(client)
```
bradrydzewski: unhandled error

drivers/azure/create.go
```diff
@@ -3,3 +3,116 @@
...
+	client.Start(ctx, p.resourceGroup, p.vmName)
```
bradrydzewski: does start return any parameters?  if it returns an error it needs to be handled

drivers/azure/create.go
```diff
@@ -3,3 +3,116 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	fmt.Printf("Starting sleep: %v\n", time.Now().Unix())
```
bradrydzewski: use logger instead of `fmt.Print` or `println`

engine/install.go
parlir: Yea I proposed that PR 😄 . I'll remove these from this PR

.gitignore
parlir: Emacs backup files

bradrydzewski: these should be removed and added to your global .gitignore file

drivers/azure/destroy.go
parlir: Yep will remove

Gopkg.toml
parlir: K will move to go.mod

kmk142789 (approved):

PR #55, opened 2019-10-10
Fix scaleway provider for all arches

tboerger:
So far it have not been possible to launch ARM or ARM64 agents, with the
update of the SDK and some additional config flags it's possible to
successfully launch ARM/ARM64 agents now.


drivers/scaleway/provider.go
```diff
@@ -49,8 +49,7 @@ func New(opts ...Option) (autoscaler.Provider, error) {
 		p.size = "dev1-l"
 	}
 	if p.image == "" {
-		// Ubuntu  for dev1-m in fra-par-1
-		p.image = "f974feac-abae-4365-b988-8ec7d1cec10d"
+		p.image = "ubuntu-bionic"
```
frebib: TIL this works. I didn't realise it would do the conversion for you. Thanks for looking into this

tboerger: It's possible that it had been added in a later version of the SDK, i had this information from the Scaleway slack channel. The create server function is using https://godoc.org/github.com/scaleway/scaleway-sdk-go/api/marketplace/v1#API.GetLocalImageIDByLabel internally, which matches the right UUID from the marketplace API.

You can fetch all available labels via `curl -s 'https://api-marketplace.scaleway.com/images?page=1&per_page=100' | jq .images[].label | tr '_' '-'`.

PR #60, opened 2020-01-07
support for custom networks in google cloud

nsigarora:
With existing code, GCP instances are not spawned if custom networks are specified for GCP instances. This PR is to add support for custom networks. Thx

### DIFF
--- a/drivers/google/provider.go
+++ b/drivers/google/provider.go
@@ -74,6 +74,9 @@ func New(opts ...Option) (autoscaler.Provider, error) {
 	if p.network == "" {
 		p.network = "global/networks/default"
 	}
+	if p.subnetwork == "" {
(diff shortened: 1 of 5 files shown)


drivers/google/provider.go
```diff
@@ -74,6 +75,9 @@ func New(opts ...Option) (autoscaler.Provider, error) {
 	if p.network == "" {
 		p.network = "global/networks/default"
 	}
+	if p.subnetwork == "" {
+		p.subnetwork = "subnetworks/default"
+	}
```
bradrydzewski: is setting a default value required?  is this a breaking change for people are using the autoscaler with google compute? 

if this has potential to break existing installations we should consider removing this code.

PR #62, opened 2020-01-29
change default interal to 1m as stated in the doc

francoispqt:
Hi,

This PR changes the default interval to 1m as stated in the doc here https://autoscale.drone.io/reference/drone-interval/

If you prefer to keep it at 5m, then the doc should be changed.

Cheers


bradrydzewski:
I intended for the docs to show how to set a custom value, not to show the default value. I see how this can be confusing and will update the docs accordingly.

PR #63, opened 2020-02-26
Add Volume IOPS configuration for AWS io1 volumes

ttousai:
This pull request implements configuring IOPS for AWS Provisioned IOPS EBS volumes (io1). 

Although autoscaler currently supports configuring EBS volume type with DRONE_AMAZON_VOLUME_TYPE, setting this parameter to `io1` causes agent creation to fail. This is because AWS requires that IOPS be set for `io1` volumes.

This PR adds a configuration `DRONE_AMAZON_VOLUME_IOPS` for configuring IOPS when using `io1` EBS volumes to fix failing agent creation.


bradrydzewski:
thanks so much!

would you mind also sending a PR to add this new parameter to our docs?
https://github.com/drone/autoscaler-docs/tree/master/content/reference

ttousai:
Yes it will be my pleasure.

PR #64, opened 2020-04-02
[GCP] Insufficient scope fix

imranismail:
This fixes an issue with GCP autoscalers not having the right scope/permission even with `Project Owner` role assigned to the service account.

### DIFF
--- a/drivers/google/provider.go
+++ b/drivers/google/provider.go
@@ -87,7 +87,7 @@ func New(opts ...Option) (autoscaler.Provider, error) {
 		p.scopes = defaultScopes
 	}
 	if p.service == nil {
-		client, err := google.DefaultClient(oauth2.NoContext, p.scopes...)
+		client, err := google.DefaultClient(oauth2.NoContext, compute.ComputeScope)
 		if err != nil {
 			return nil, err
 		}


imranismail:
> This doesn't fix the scopes, this just overrides the scopes. The default scope should be fixed properly.

I'm sorry, where is the default scoped defined?

From GCP go sdk examples, I see that this is the required scope for creating compute instances.

tboerger:
You have edited the file where the default is defined: https://github.com/imranismail/autoscaler/blob/fc0727b709406c88c416254503a3320837ef126a/drivers/google/provider.go#L28

bradrydzewski:
I think @imranismail is correct.

The default scopes are assigned to the instance:
https://github.com/drone/autoscaler/blob/master/drivers/google/create.go#L93

The `compute.Compute` scopes should probably be used when creating the client.  The docker machine library takes the same approach:
https://github.com/docker/machine/blob/fb59f089f10dabb90e45491999ec2589513a6980/drivers/google/compute_util.go#L53

bradrydzewski:
Yes, as I understand it, we do need two sets of scopes. One scope is for the client (e.g. create, terminate instances).  The second set of scopes are assigned to created instances.  We would not want instances created by the autoscaler to have the same scope as the autoscaler, otherwise a malicious would have read/write access to the gcp account.

imranismail:
> Yes, as I understand it, we do need two sets of scopes. One scope is for the client (e.g. create, terminate instances) and should be hardcoded. The second set of scopes are optional and are assigned to instances that the autoscaler creates. We would not want instances created by the autoscaler to have the same scope as the autoscaler itself, since the autoscaler has read/write access. I admit I am not very familiar with gcp, however, it should behave in a manner similar to docker machine.

This is correct,

The agent VM that autoscaler creates doesn't need the same permission.

For those looking to get this to work on GCP.

1. Autoscaler needs the right scope requested (fixed with this PR)
2. The machine Autoscaler runs in needs the right scope. In my case I was running it in GKE, I had to create a special node pool with the `compute-rw` scope (https://cloud.google.com/sdk/gcloud/reference/container/clusters/create#--scopes)
3. Autoscaler needs the right service account (I created one with Compute Admin role and mounted it into the autoscaler pod)

PR #68, opened 2020-07-09
Configure Service Account name for GCP instances

ademariag:
Adds option for passing an alternative GCP Service Account using the DRONE_GOOGLE_SERVICE_ACCOUNT_EMAIL env variable

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -152,7 +152,7 @@ type (
 			Network             string            `envconfig:"DRONE_GOOGLE_NETWORK"`
 			Subnetwork          string            `envconfig:"DRONE_GOOGLE_SUBNETWORK"`
 			Labels              map[string]string `envconfig:"DRONE_GOOGLE_LABELS"`
-			Scopes       []string          `envconfig:"DRONE_GOOGLE_SCOPES"`
+			Scopes              []string          `envconfig:"DRONE_GOOGLE_SCOPES"`
 			ServiceAccountEmail string            `envconfig:"DRONE_GOOGLE_SERVICE_ACCOUNT_EMAIL"`
 			DiskSize            int64             `envconfig:"DRONE_GOOGLE_DISK_SIZE"`
 			DiskType            string            `envconfig:"DRONE_GOOGLE_DISK_TYPE"`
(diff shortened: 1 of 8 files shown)


bradrydzewski:
LGTM

config/config.go
```diff
@@ -147,20 +147,21 @@ type (
...
+			MachineImage        string            `envconfig:"DRONE_GOOGLE_MACHINE_IMAGE"`
+			Network             string            `envconfig:"DRONE_GOOGLE_NETWORK"`
+			Subnetwork          string            `envconfig:"DRONE_GOOGLE_SUBNETWORK"`
+			Labels              map[string]string `envconfig:"DRONE_GOOGLE_LABELS"`
+			Scopes              string            `envconfig:"DRONE_GOOGLE_SCOPES"`
+			ServiceAccountEmail string            `envconfig:"DRONE_GOOGLE_SERVICE_ACCOUNT_EMAIL"`
+			DiskSize            int64             `envconfig:"DRONE_GOOGLE_DISK_SIZE"`
+			DiskType            string            `envconfig:"DRONE_GOOGLE_DISK_TYPE"`
+			Project             string            `envconfig:"DRONE_GOOGLE_PROJECT"`
+      PrivateIP           bool              `split_words:"true"`
```
ashwilliams1, line 160: looks like an alignment issue. Go uses tabs instead of spaces. I recommend using the `gofmt` [1] utility to automatically format the code, which will ensure it conforms to style guides. If you are using an edit (like vscode) there is an official Go plugin that will automatically format when you save the file.

[1] https://blog.golang.org/gofmt

drivers/google/create.go
```diff
@@ -135,6 +135,7 @@ func (p *provider) Create(ctx context.Context, opts autoscaler.InstanceCreateOpt
 		Region:   p.zone,
 		Size:     p.size,
 		Address:  resp.NetworkInterfaces[0].AccessConfigs[0].NatIP,
+		ServiceAccountEmail: p.serviceAccountEmail,
```
ashwilliams1, line 138: please use `gofmt` here too :)

ashwilliams1:
just need gofmt, but then can merge! thanks @ademariag for the improvements!

PR #69, opened 2020-07-09
Adds missing WithScopes function

ademariag:
The current SCOPES are not being applied because of a missing setup function.

### DIFF
--- a/drivers/google/option.go
+++ b/drivers/google/option.go
@@ -7,7 +7,6 @@ package google
 import (
 	"io/ioutil"
 	"net/http"
-	"strings"
 
 	"github.com/drone/autoscaler/drivers/internal/userdata"
 
@@ -132,8 +131,8 @@ func WithZone(zone string) Option {
 }
 
 // WithScopes returns an option to set the scopes.
-func WithScopes(scopes string) Option {
+func WithScopes(scopes ...string) Option {
 	return func(p *provider) {
-		p.scopes = strings.Split(scopes, ",")
+		p.scopes = scopes
 	}
 }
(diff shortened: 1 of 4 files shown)


drivers/google/option_test.go
```diff
@@ -51,4 +54,7 @@ func TestOptions(t *testing.T) {
 	if got, want := p.zone, "us-central1-f"; got != want {
 		t.Errorf("Want zone %q, got %q", want, got)
 	}
+	if got, want := p.scopes, strings.Split("scope1,scope2", ","); ! cmp.Equal(got, want) {
```
bradrydzewski, line 57: I recommend changing scopes to a `[]string` in the `Config` struct.  The library we use supports parsing comma-separated environment variables into an array automatically.
https://github.com/drone/autoscaler/blob/master/config/config.go#L154

drivers/google/option.go
```diff
@@ -5,11 +5,10 @@
 package google
 
 import (
+	"github.com/drone/autoscaler/drivers/internal/userdata"
```
ashwilliams1, line 8: hey there, I recommend reverting back.  The Go style guide recommends that imports should be grouped by standard library first, then application-specific imports, and then third party imports.

ademariag: Sorry for once I relied to use Goland but I messed things up rebasing. Fixing it now :)

ademariag: Here, sorry. I don't know how I got myself into that state but all sorted.

bradrydzewski:
thanks!

PR #72, opened 2020-08-26
Negotiate API Version

donny-dont:
The `docker.NewClient` function is deprecated for `docker.NewClientWithOpts`. Specify the already present options but also use API negotiation rather than using the explicit default version.

### DIFF
--- a/engine/docker.go
+++ b/engine/docker.go
@@ -11,7 +11,6 @@ import (
 	"io"
 	"net/http"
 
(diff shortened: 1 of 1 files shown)


donny-dont:
@bradrydzewski so either we can use this or use the `FromEnv` like what you are doing at https://github.com/drone-runners/drone-runner-docker/blob/master/engine/engine.go#L47 just let me know your preference.

PR #76, opened 2020-09-15
Add Max Procs Option

swjclarke:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -219,5 +219,6 @@ type (
 		Devices    string
 		Privileged string
 		EnvFile    string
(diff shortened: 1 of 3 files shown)


ashwilliams1:
FYI you can use this setting to pass additional parameters to the agent:
https://autoscale.drone.io/reference/drone-agent-environ/

For example:
```
DRONE_AGENT_ENVIRON=DRONE_RUNNER_MAX_PROCS:2
```

swjclarke:
Awesome will use that for testing, will keep this open as a we should mirror runner options to prevent that being a massive list.

PR #77, opened 2020-10-11
Allow autoscaler to work with pre-existing static builders.

techknowlogick:
Use case: I have several reserved instances on equinix metal (nee Packet), but have credits on DigitalOcean that I use for overflow capacity. Before this PR the autoscaler will overprovision, thankfully it doesn't impact the bill too much, but this PR would decrease our bill. In theory, this could allow multiple instances of the autoscaler to run and have for ex 1-3 instances to run on provider 1, 4-5 with provider 2, and anymore on provider 3 (way more complex use cases would also be possible).

Hopefully you find this PR useful, but if not then that's ok too as I could still use it for myself, but figured I'd contribute it so others can use it as well.

Signed-off-by: Matti R <matti@mdranta.net>


tboerger (approved):
IMHO this makes sense

config/config.go
```diff
@@ -32,9 +32,10 @@ type (
 		}
 
 		Pool struct {
-			Min    int           `default:"2"`
-			Max    int           `default:"4"`
-			MinAge time.Duration `default:"55m" split_words:"true"`
+			Min              int           `default:"2"`
+			Max              int           `default:"4"`
+			MinAge           time.Duration `default:"55m" split_words:"true"`
+			ExistingCapacity int           `envconfig:"DRONE_POOL_EXISTING_CAPACITY" default:"0"`
```
tboerger, line 38: You should use `split_words` instead of enforcing the env variable:

```diff
- 			ExistingCapacity int           `envconfig:"DRONE_POOL_EXISTING_CAPACITY" default:"0"`
+ 			ExistingCapacity int           `default:"0" split_words:"true"`
```

engine/planner.go
```diff
@@ -151,7 +153,7 @@ func (p *planner) mark(ctx context.Context, n int) error {
 		logger.WithField("servers-to-terminate", n).
 			WithField("servers-running", len(servers)).
 			WithField("min-pool", p.min).
-			Debugf("abort terminating %d instances to ensure minimum capacity met")
+			Debugf("abort terminating %d instances to ensure minimum capacity met", n)
```
tboerger, line 156: As it already got the field `servers-to-terminate` I would just rephrase the message and remove the parameter.

techknowlogick: Thanks. I've just updated.

config/config.go
techknowlogick: Thanks. I've just updated.

PR #78, opened 2020-10-13
Update to latest Equinix Metal (nee Packet) library which has updated endpoints

techknowlogick:
Now that Packet is Equinix Metal, they have updated their API URLs (not sure when the old ones expire, but probably better to update sooner so that there would be no interruption of service)


techknowlogick:
Hmm... Updated packet client uses https://github.com/hashicorp/go-retryablehttp which isn't intercepted by gock.

techknowlogick:
Ah, excellent that works now :)

PR #81, opened 2020-11-29
Reaper and Pinger configuration

ttousai:
This PR makes the following changes:

1. Use the DRONE_REAPER_ENABLED config under config.Config as the reaper feature flag instead of DRONE_ENABLE_REAPER.
1. Add a Pinger struct under config.Config to hold the pinger feature flag and pinger interval.
1. Use the new DRONE_PINGER_ENABLED config as the pinger feature flag instead of DRONE_ENABLE_PINGER.


bradrydzewski:
:wave: thanks! 

I think we should still provide backward compatibility for _DRONE_ENABLE_PINGER_ and )DRONE_ENABLE_REAPER_.  We handle this scenario in some of our other repositories by adding an init function that looks for legacy environment variables, and if it finds them, sets the new environment variables.

This function would go in `config/load.go`

```Go
// legacy environment variables. the key is the legacy
// variable name, and the value is the new variable name.
var legacy = map[string]string{
	"DRONE_ENABLE_PINGER": "DRONE_PINGER_ENABLED",
	"DRONE_ENABLE_REAPER": "DRONE_REAPER_ENABLED",
}

func init() {
	// loop through legacy environment variable and, if set
	// rewrite to the new variable name.
	for k, v := range legacy {
		if s, ok := os.LookupEnv(k); ok {
			os.Setenv(v, s)
		}
	}
}
```

tboerger (approved):

PR #83, opened 2021-01-29
Add the ability to configure agent logging

mach6:
Adds the ability to specify the logging driver and
options for the runner (agent), gc, and watchtower
containers.

Example config;
```
DRONE_AGENT_LOGGING_DRIVER=json-file
DRONE_AGENT_LOGGING_OPTIONS=max-size:100m,max-file:5
```
ref: https://docs.docker.com/config/containers/logging/configure/#supported-logging-drivers


engine/install.go
```diff
@@ -371,13 +398,28 @@ func (i *installer) setupGarbageCollector(ctx context.Context, client docker.API
...
 	}
 	return client.ContainerStart(ctx, res.ID, types.ContainerStartOptions{})
 }
 
+func (i *installer) getDaemonLoggingDriver(ctx context.Context, client docker.APIClient) string {
+	logger := logger.FromContext(ctx)
+
+	info, err := client.Info(ctx)
+	if err != nil {
+		logger.Warnln("cannot acquire logging driver, using 'json-file'")
```
mach6, line 417: hmm.. perhaps I should not ignore the `err` here.. will fix!

mach6:
@bradrydzewski @tboerger Any feedback on this one? Cheers.

tboerger:
Why don't you just define the logging driver on Docker daemon level? IMHO it doesn't make sense to define that per container.

mach6:
> Why don't you just define the logging driver on Docker daemon level? IMHO it doesn't make sense to define that per container.

While one can certainly do this; per container config enables finer configuration — for example; if one does not want the drone-gc or watchtower logging behavior to be the same  as the agent.. maybe one wants the agents logs to emit to via the gcplogs driver and the drone-gc to emit via json-file, for example.

mach6:
@tboerger @bradrydzewski any additional thoughts on this one?

mach6:
PR re-based against v1.8.1 and force-pushed

kmk142789 (approved):

PR #85, opened 2021-04-30
Add ability to spawn runners in DO vpc

tomjohnburton:


.drone.yml
```diff
@@ -57,7 +57,9 @@ steps:
   image: plugins/docker
   settings:
     auto_tag: true
-    repo: drone/autoscaler
+    repo: ghcr.io/chainflip-io/autoscaler
```
tboerger, line 60: This should not be part of the pr.

.gitignore
```diff
@@ -7,4 +7,5 @@ vendor
 *.bak
 *.out
 *.db
-*.env
\ No newline at end of file
+*.env
+.idea
```
tboerger, line 11: Ignores for your dev environment should not be part of the project ignore.

config/config.go
```diff
@@ -141,6 +141,7 @@ type (
 			SSHKey       string
 			Size         string
 			Firewall     string
+			Vpc     string
```
tboerger, line 144: Wrong formatting

drivers/digitalocean/provider.go
```diff
@@ -25,6 +25,7 @@ type provider struct {
 	size      string
 	image     string
 	firewall  string
+	vpc		  string
```
tboerger, line 28: Also wrong formatting

tboerger (commented):
I have added some comments, but beside that the vpc option have not been added to the config struct.

tomjohnburton:
Go into a bit of a death spiral with the PR with some broken tests. Restarted with this one #86

PR #91, opened 2021-05-31
Allow setting a name prefix for agents

julienduchesne:
This will allows us to use `drone-{id}` instead of `agent-{id}`


config/load_test.go
```diff
@@ -215,7 +215,8 @@ var jsonConfig = []byte(`{
     "Image": "drone/drone-runner-docker:latest",
     "Concurrency": 2,
     "KeepaliveTime": 360000000000,
-    "KeepaliveTimeout": 30000000000
+    "KeepaliveTimeout": 30000000000,
+	"NamePrefix": "agent-"
```
inkel, line 219: You might want to run `go fmt` on this file.

engine/planner.go
```diff
@@ -108,9 +109,13 @@ func (p *planner) alloc(ctx context.Context, n int) error {
 	logger := logger.FromContext(ctx)
 	logger.Debugf("allocate %d servers", n)
 
+	namePrefix := p.namePrefix
+	if namePrefix == "" {
+		namePrefix = "agent-"
+	}
```
jvrplmlmn, line 115: Is this block necessary given that we default to `agent-`?

```go
NamePrefix string `envconfig:"DRONE_AGENT_NAME_PREFIX" default:"agent-"`
```

Alternatively, would it be also necessary to have this in `server/servers.go`?

julienduchesne: `server/servers.go` uses the config directly which has a default so it's not needed while the planner is a separate struct. It doesn't necessarily draw from the config. At least that's my iterpretation

inkel: Same interpretation here.

PR #95, opened 2021-06-28
Enable Reaper from main config

julienduchesne:
There are currently two flags for the reaper:
- DRONE_ENABLE_REAPER
- DRONE_REAPER_ENABLED

This removes the DRONE_ENABLE_REAPER flag which doesn't follow the regular pattern


bradrydzewski:
added via #81

PR #96, opened 2021-07-13
ignore error servers determining capacity, terminated aged error servers

herrgruebner:
This PR does two things.

1) Changes the capacity calculation so that servers in the error state no longer count for capacity

Servers in the error state can't build anything, so factoring them in to capacity calculations gives an inaccurate view of capacity.

2) Terminates errored servers after they've reached the minimum age.

At the moment, servers in the error state do not get terminated automatically, or preferentially over running servers. If they do get terminated, it will just be because they happen to be in the right spot on the idle server list.

In this PR, I've gone with the approach of automatically terminating ones that have aged out. An alternative might be preferred where errored servers are terminated first instead of healthy servers.

I think this is a sensible approach because it allows some time to fix servers in the error state, but also doesn't keep broken servers sitting around costing money.

I've added some tests, but I'll freely admit my Go is terrible so I'm happy to make changes.

### DIFF
--- a/engine/planner.go
+++ b/engine/planner.go
@@ -81,7 +81,7 @@ func (p *planner) Plan(ctx context.Context) error {
 	// if the server differential to handle the build volume
 	// is positive, we can reduce server capacity.
 	if diff < 0 {
-		return p.mark(ctx,
+		return p.markForTermination(ctx,
 			// we should adjust the desired capacity to ensure
 			// we maintain the minimum required server count.
 			serverFloor(servers, abs(diff), p.min),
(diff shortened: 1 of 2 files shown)


bradrydzewski:
while I appreciate the pull request, we want to avoid infinite loops of creating and tearing down instances which can have very real and negative impacts. See <https://github.com/drone/autoscaler/pull/51#issuecomment-531580254>.  If you are comfortable automatically tearing down errored instances you can set `DRONE_ENABLE_REAPER=true` which will run hourly and tear down errored instances.

PR #97, opened 2021-07-15
Add option to specify Amazon EC2 instance metadata accessiblity

hhamalai:
Adds an option to set instance metadata service (IMDS) access option to require the usage of tokens (IMDSv2). The default value, "optional", will not change the current behavior.

### DIFF
--- a/drivers/amazon/provider.go
+++ b/drivers/amazon/provider.go
@@ -36,6 +36,7 @@ type provider struct {
 	tags          map[string]string
 	iamProfileArn string
 	spotInstance  bool
+	imdsTokens    string
 }
 
 func (p *provider) getClient() *ec2.EC2 {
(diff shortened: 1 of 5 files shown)


drivers/amazon/provider.go
```diff
@@ -78,5 +79,8 @@ func New(opts ...Option) autoscaler.Provider {
 	if p.userdata == nil {
 		p.userdata = userdata.T
 	}
+	if p.imdsTokens == "" {
+		p.imdsTokens = "optional"
+	}
```
bradrydzewski, line 84: I recommend removing this change, per above comments about not automatically setting a value (which otherwise needs to be fully tested for regressions through setting up an environment, etc)

bradrydzewski (changes requested):

hhamalai:
Thanks for the comments @bradrydzewski, I carried out the changes you suggested and rebased on master.

hhamalai:
@bradrydzewski , any ideas to improve this PR, or is it off the scope for the project, and should I close it and try to tackle this issue with another approach?

naikrovek:
I am struggling to understand how this is still a PR and hasn't been merged in the three years that it has been alive.

I know what you're going to say:

> Have you tried Harness.io, the modern software delivery platform?

Quiet.  Don't forsake security in favor of upselling.

PR #101, opened 2021-09-19
GCP multi-zone support

TheJokersThief:
* Adds support for autoscaling in multiple zones in one region on GCP.
* If a region is specified via `DRONE_GOOGLE_REGION`, it will randomly choose a zone within that region and ignore the zone specified in `DRONE_GOOGLE_ZONE`
* When destroying an instance, the destroy method will use the `Region` (which is actually a zone ID)

Quick question: I wasn't totally sure on the method for adding and recognising an optional environment variable, so please let me know if I missed something there :)


drivers/google/destroy.go
```diff
@@ -13,7 +13,11 @@ import (
 )
 
 func (p *provider) Destroy(ctx context.Context, instance *autoscaler.Instance) error {
-	op, err := p.service.Instances.Delete(p.project, p.zone, instance.ID).Context(ctx).Do()
+	zone := p.zone
+	if p.region != "" {
+		zone = p.region	
+	}
```
bradrydzewski, line 19: @TheJokersThief I made a minor change to preserve the existing behavior.  The google docs did not mention that region could be used, and I don't have enough experience with Google Cloud to confidently merge the pull request without spinning up a full environment and running some tests (which would be time consuming).  So in the interest of getting this merged quickly, I figured maybe we could implement the above logic as a compromise? Look ok to you?

bradrydzewski:
lgtm pending answer to my question

bradrydzewski (approved):

drivers/google/create.go
```diff
@@ -49,6 +50,19 @@ func (p *provider) Create(ctx context.Context, opts autoscaler.InstanceCreateOpt
...
+		// If region was specified, use region to choose a random zone
+		region, err := p.service.Regions.Get(p.project, p.region).Do()
+		if err != nil {
+			logger.WithError(err).
+				Errorln("cannot get region details")
+			return nil, err
+		}
+		randomZoneIndex := rand.Intn(len(region.Zones))
+		zoneSelfLink := region.Zones[randomZoneIndex]
+		p.zone = zoneSelfLink[strings.LastIndex(zoneSelfLink, "/")+1:]
```
bradrydzewski, line 63: you could probably simplify as `p.zone = path.Base(zoneSelfLink)`

bradrydzewski: sorry, one gotcha here is that `p.zone` is not going to be thread safe to write to.  do we need to overwrite `p.zone` or can this be stored in a local variable?

drivers/google/destroy.go
bradrydzewski: also does `waitZoneOperation` work with a region?

bradrydzewski: also just want to confirm that if we pass `region` to the `Delete` function, will this have any impact on the subsequent `waitZoneOperation`?  Sorry, I'm not too familiar with gcp.

drivers/google/create.go
TheJokersThief: You are right on both accounts - thanks :) This also uncovered that the operation wait function relied on having `p.zone` and succeeded during tests because I was obviously running that one test synchronously 😅 So I've changed that function to take a `zone` as an argument.

bradrydzewski:
@TheJokersThief thanks, the code looks great.  I do want to ask a few design questions before we merge since there are a few different ways we could solve this problem.  Please note that these are just probing questions, I'm not advocating for or suggesting any changes, but want to make sure we consider a few different options before we finalize this.

1. do we want to make the API call to get the list of zones every time the install function is called, or is this something we would want to execute once and then cache?
2. if we want to make the API call every time, if the API call fails we currently error.  Instead of returning an error we could write the error message to the logs, but then fallback to the default zone.  This might make it slightly more resilient in the event the Google API is misbehaving (although not sure this is a problem in practice)
3. is it possible that someone would want to use a subset of zones in a region?  If yes, perhaps instead of providing a region and dynamically querying region for a list of zones, we allow the end-user to configure one or many zones instead of a single zone.

TheJokersThief:
That's a really good point 🤔 Yep, I think having someone supply their list of zones (option 3) would be a good idea :) I can't think of a scenario where someone would want to use just a subset of zones, but it does feel like the option that would give users the most control.

Because `DRONE_GOOGLE_ZONE` is a variable, it would probably need to be a string of comma-separated values, right?

Also, should it be renamed `[...]_ZONES`? We could piggyback off this to support the singular name for backwards compatibility: https://github.com/drone/autoscaler/blob/master/config/load.go#L18-L21

bradrydzewski:
> it would probably need to be a string of comma-separated values, right?

Yes, this would be automatically supported by the library we use to decode environment variable into the structure.  So we can change from a string to string slice.

```diff
-Zone                string            `envconfig:"DRONE_GOOGLE_ZONE"`
+Zone                []string          `envconfig:"DRONE_GOOGLE_ZONE"`
```

This would not have any adverse impact on existing users since a string value would be interpreted as a single value in a comma-separated array. Both of the below examples would decode into a string slice:

```text
DRONE_GOOGLE_ZONE=us-central1-a
DRONE_GOOGLE_ZONE=us-central1-a,us-central1-b
```

drivers/google/option.go
```diff
@@ -134,9 +134,16 @@ func WithUserDataKey(text string) Option {
 }
 
 // WithZone returns an option to set the target zone.
-func WithZone(zone string) Option {
+func WithZones(zones []string) Option {
```
bradrydzewski, line 137: one minor change for consistency ... can you do `WithZones(zones ...string)` similar to this:
https://github.com/drone/autoscaler/blob/350f5a87a679c5797f9d35a1b448362843cdcf5f/drivers/google/option.go#L96

drivers/google/option.go
```diff
@@ -134,9 +134,16 @@ func WithUserDataKey(text string) Option {
...
 	return func(p *provider) {
-		p.zone = zone
+		p.zones = zones
+	}
+}
+
+// WithZone returns an option to set the target zone.
+func WithRegion(region string) Option {
```
bradrydzewski, line 144: can we remove?

bradrydzewski:
@TheJokersThief I've added two minor comments and then we should be good to merge.  Thanks for your patience through the review process!

drivers/google/option.go
TheJokersThief: Yep and removed :)

drivers/google/option.go
TheJokersThief: Got it, thanks 😄

TheJokersThief:
@bradrydzewski No worries at all :) Thanks for working through it with me! I look forward to whenever this can be merged, it'll help Drone's GCP story quite a bit.

bradrydzewski:
thanks!  I'll give it a final sanity check tomorrow and should have it merged this weekend.

bradrydzewski (approved):

bradrydzewski:
@TheJokersThief this looks good and is ready to merge.  I may want to tag a release from master before I merge, just in case someone needs to revert back for some unknown reason (sorry, I am sometimes a bit overly cautious).  I have to write up the release notes which will take some time since we merged a ton of pull requests last week, but once complete, we can merge and cut a new minor release with your changes.

PR #102, opened 2021-10-06
pinger: Log when pinging fails

iainlane:
Send a debug log when we fail & retry, and a warning when we're about to
try to delete a server


eoinmcafee00:
Hey @iainlane 

This was the ticket i meant to close due to Brad adding this a while back https://github.com/drone/autoscaler/commit/e60ca12937f6eca6221f6ae10d4ce5147ce373c7

Cheers
Eoin

PR #104, opened 2021-10-27
Add template functions for userdata

iainlane:
It's good to have some common useful functions available. In particular I wanted to use `nindent` to include a via `write_files`.


iainlane:
Alright, I've done as requested. This depends on https://github.com/drone/funcmap/pull/2.

Note that I need to have `Funcs` rather than `SafeFuncs` as we're fetching from the environment and `env` is not safe by funcmap's definition.

go.mod
```diff
@@ -68,3 +76,5 @@ require (
 	gotest.tools v2.2.0+incompatible // indirect
 	honnef.co/go/tools v0.0.0-20190523083050-ea95bdfd59fc // indirect
 )
+
+replace github.com/drone/funcmap => github.com/iainlane/funcmap v0.0.0-20211116113722-13f662008062
```
iainlane, line 80: this needs fixing once https://github.com/drone/funcmap/pull/2 is merged

PR #115, opened 2022-04-04
Enforce and configure an API limit for GCP

TheJokersThief:
## The Problem

When the autoscaler spins up a large number of agents in a short period of time in GCP, it starts checking if the creation operation finished for each one. The larger the scale, the higher the burst. This can lead to being ratelimited as you're allowed [a max of 25/s or 1500/min read requests to the API](https://cloud.google.com/compute/docs/api-rate-limits).

When the autoscaler gets rate limited, it errors and loses track of agents, which can leave them orphaned.

## A Proposed Solution

Make a configurable _internal_ rate limit for the google provider that lets us set an acceptable max rate for requests.

This way, we can also play nice with other services that use the same API in the same project, and people can choose their rate to achieve that balance.


bradrydzewski:
thanks, look great ... I had never seen the ratelimit package before so nice to learn something new :) 

as an aside, I think we might need to look at the use of context in the wait operations.  If the program is gracefully terminated the context _might_ get cancelled which could prematurely cancel a wait operation.  I will do some more exploration, but wanted to bring that up while we are looking at that portion of the code.  I will handle in a separate patch.

```
op, err := p.service.GlobalOperations.Get(p.project, name).Context(ctx).Do()
```

PR #119, opened 2022-08-15
Set us-west-2 agent ami to 18.04 to avoid deprecation

sawp-d:
Change the default AWS AMI image for agents from 16.04 to 18.04 to avoid the deprecation notice.


tphoney:
Hey @sawp-d 

```
"us-west-2": "ami-0aab355e1bfa1e72e"
```

has been updated on master, thanks for the PR !

PR #124, opened 2023-01-06
Support labels in Hetzner cloud

gzm0:
The current version of the Hetzner Go library (1.4.0) does not have support for labels.

We upgrade the Hetzner Go library to the latest version that still supports Go 1.12 (1.17.0).


gzm0:
Rebased and fixed conflicts after #126.

gzm0:
Seems like the tests fail because the newer Hetzner libraries expect more precise responses from the REST APIs. Maybe we can get away with mocking a level lower (i.e. just the client, not the actual rest endpoints).

I'll try to give this a shot.

PR #125, opened 2023-01-12
Feature: support minimum idle time before terminating server

tjstansell:
This attempts to add support for `DRONE_POOL_MIN_IDLE`, which will define the minimum idle time before a server is terminated.  It defaults to `0`, but if you set it to `30m` then it will only terminate a server after it has sat idle for at least 30 minutes (no jobs have run).  It utilizes the `Updated` server property, which I believe gets updated as the server changes state from `running` to `pending` status.  As it completes the builds, this should represent when the last build finished.  As such, if jobs continue to fire regularly, then it stays online, but if jobs stop, it would then terminate after this duration.

I'm currently trying to get a test installation online to test these changes, but figured I'd see if folks had feedback regardless.

### DIFF
--- a/engine/planner.go
+++ b/engine/planner.go
@@ -186,7 +186,7 @@ func (p *planner) mark(ctx context.Context, n int) error {
 			continue
 		}
 
-		// skipi servers that have not reached a min idle time
+		// skip servers that have not reached a min idle time
 		if time.Now().Before(time.Unix(server.Updated, 0).Add(p.tti)) {
 			logger.
 				WithField("server", server.Name).
--- a/engine/planner.go
+++ b/engine/planner.go
@@ -186,6 +186,16 @@ func (p *planner) mark(ctx context.Context, n int) error {
 			continue
 		}
 
+		// skipi servers that have not reached a min idle time
+		if time.Now().Before(time.Unix(server.Updated, 0).Add(p.tti)) {
+			logger.
+				WithField("server", server.Name).
+				WithField("idle", timeDiff(time.Now(), time.Unix(server.Updated, 0))).
+				WithField("min-idle", p.tti).
+				Debugln("server min-idle not reached")
+			continue
+		}
+
 		idle = append(idle, server)
 		logger.WithField("server", server.Name).
 			Debugln("server is idle")
--- a/engine/planner.go
+++ b/engine/planner.go
@@ -30,6 +30,7 @@ type planner struct {
 	cap        int           // capacity per-server
 	buffer     int           // buffer capacity to have warm and ready
 	ttu        time.Duration // minimum server age
+	tti        time.Duration // minimum server idle time
 	labels     map[string]string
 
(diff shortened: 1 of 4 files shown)


tjstansell:
Hm... the above changes cause slack notifications to be sent since they trigger on Updates.

bradrydzewski:
@tjstansell conceptually I like the idea, however, the `Updated` date does not reflect the time the last build was executed on the machine, it just reflects the last time the database entry was updated. We do not currently store the last build date in the autoscaler database.  I am not sure how to achieve this at the moment -- would need to think on this more.

bradrydzewski:
ok, reading the patch, I think I have a better understanding.  It looks like a forced save is used to update the `Updated` date for the server.  I would probably recommend we create a dedicated field to store this data.  The main reason is that I cannot guarantee the `Updated` data would not be updated in the future, for other reasons, and not break your logic.

engine/planner.go
```diff
@@ -76,6 +77,13 @@ func (p *planner) Plan(ctx context.Context) error {
 
 	ctx = logger.WithContext(ctx, log)
 
+	_, err = p.updateBusy(ctx)
```
bradrydzewski, line 80: I would recommend only executing this routine if the user has enabled this setting.  This avoids unnecessary database and API calls

```
if p.MinIdle > 0 {
  _, err = p.updateBusy(ctx)
}
```

tjstansell: ah yes.  i remember thinking that at one point before falling asleep.  i'll update it.

tjstansell:
Thanks for all the feedback so far.  I updated it to use a new LastBusy field to track this.  Hopefully I did it right. :)  Let me know if we need anything else.

tjstansell:
Actually, I think I now see how the db migration bits work ... working on updating it to do a proper migration.

tjstansell:
I deployed this to our internal instance and it seems to be working!  Let me know if there's anything you'd like to see changed, though... Thanks!

tjstansell:
Any additional thoughts on this?  Thanks.

tjstansell:
@bradrydzewski curious if there's any way to get some progress on this so it doesn't just get lost.  fwiw, we've been running on this code on our production instances for months now without any issue.

tjstansell:
@bradrydzewski sorry to keep pestering, but I've been trying to get some progress on this PR now for over 3 months with zero responses.  Is there any chance someone can take a look at this before it's so stale it's no longer valid?

alikhil:
It would be nice to have this feature merged

tjstansell:
FWIW, we've been running this patched version for the last year and a half and have not noticed any issues... I did just rebase my patch against the current master branch to pick up the other features that have been added, so this is now current.  I'd love it if someone could evaluate this and hopefully merge it this time...

oscarcipriano2000:
is possible to merge this pr?

PR #133, opened 2023-04-03
fixes issue in idle termination logic

eoinmcafee00:
Issue raised where 'pending' state wasn't taken into consideration as 'busy'.

### DIFF
--- a/engine/planner.go
+++ b/engine/planner.go
@@ -192,7 +192,7 @@ func (p *planner) mark(ctx context.Context, n int) error {
 	}
 
 	// if there are no idle servers, there are no servers
-	// to retire and we can exit.
+	// to retire, we can exit.
 	if len(idle) == 0 {
 		logger.Debugln("no idle servers to shutdown")
 	}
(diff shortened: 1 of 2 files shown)


engine/planner.go
```diff
@@ -265,7 +265,7 @@ func (p *planner) listBusy(ctx context.Context) (map[string]struct{}, error) {
 		if p.match(stage) == false {
 			continue
 		}
-		if stage.Status == drone.StatusRunning {
+		if stage.Status == drone.StatusRunning || stage.Status == drone.StatusPending {
```
eoinmcafee00, line 268: @bradrydzewski - out of the other statuses do you think it maybe should be:

```
if stage.Status == drone.StatusRunning || stage.Status == drone.StatusPending ||
			stage.Status == drone.StatusBlocked || stage.Status == drone.StatusWaiting {
			busy[stage.Machine] = struct{}{}
		}
```

vistaarjuneja (approved):

PR #144, opened 2024-08-28
fix: [CI-14032]: Add option to specify Amazon EC2 instance metadata a…

ShobhitSingh11:
…ccessiblity

### DIFF
--- a/drivers/amazon/create.go
+++ b/drivers/amazon/create.go
@@ -139,6 +139,12 @@ func (p *provider) create(ctx context.Context, opts autoscaler.InstanceCreateOpt
 		},
 	}
 
+	if p.imdsTokens != "" {
+		in.MetadataOptions = &ec2.InstanceMetadataOptionsRequest{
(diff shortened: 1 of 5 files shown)


vistaarjuneja (approved):

drivers/amazon/create.go
```diff
@@ -139,6 +139,12 @@ func (p *provider) create(ctx context.Context, opts autoscaler.InstanceCreateOpt
 		},
 	}
 
+	if p.imdsTokens != "" {
```
vistaarjuneja, line 142: Nit: use the same convention as others instead of directly modifying in.

```
var in *ec2.InstanceMetadataOptionsRequest

if p.imdsTokens != "" {
    metadataOptions = &ec2.InstanceMetadataOptionsRequest{
          HttpTokens: aws.String(p.imdsTokens),
    }
}

```

vistaarjuneja (approved):

